PR #1, opened 2022-09-19
Implement swift functions in Bazel [BUILD-317]

krisukox:
This PR adds `swift_add_*` functions to Bazel.
The main purpose is to use a consistent set of compiler options.

Test PRs:
https://github.com/swift-nav/gnss-converters-private/pull/1324
https://github.com/swift-nav/libswiftnav-private/pull/391


swift_targets.bzl
```diff
@@ -0,0 +1,251 @@
...
+            fail("Do not specify -frtti directly, use 'rtti' argument instead")
+        if flag == "-fno-rtti":
+            fail("Do not specify -fno-rtti directly, avoid 'rtti' argument instead")
+        if flag == "-Wno-error":
+            fail("Do not specify -Wno-error directly, use 'warning' to disable -Werror")
+
+    all_flags = []
+
+    # if not warning:
+    #     all_flags += ["-Werror", "-Wno-error=deprecated-declarations"]
```
krisukox, line 29: I had to disable `-Werror` because compiler complains when using C flag for C++ and vice versa.

swift_targets.bzl
```diff
@@ -0,0 +1,233 @@
...
+        "-Wformat",
+        "-Wformat-security",
+        "-Wformat-y2k",
+        "-Wimplicit-fallthrough",
+        "-Wimport",
+        "-Winit-self",
+        "-Winvalid-pch",
+        "-Wmissing-braces",
+        "-Wmissing-field-initializers",
+        # "-Wmissing-include-dirs",
```
krisukox, line 50: I commented out `-Wmissing-include-dirs` because gcc-11 (this is not the case with clang-14) complains when e.g. `include` directory "inherited" from the `swiftnav` library is not valid for the `ixcom` library.

I tried to workaround it with the `-I` flag but I couldn't find a solution. We should probably create a ticket about it.

PR #2, opened 2022-09-29
Add Rules For Clang-Format [BUILD-305]

krisukox:
This PR adds rules for clang-format.

`clang-format-check` is implemented using the actions API - we can use cache functionality.
In order to check formatting in a repo use: `bazel build //... --config clang-format-check`
In order to format a repo use: `bazel run //bazel/clang_format`

explanation:
`clang-format-check` is defined in the `.bazelrc` file. Bazel passes all targets (`//...`) to the `clang_format/clang_format_check.bzl:clang_format_check_aspect` aspect. Bazel then iterates over files in the target and runs `clang-format` on it.
The clang format version is searched in the `clang_format/choose_clang_format.bzl` file. The clang format version is searched only once, and then we use it from the cache.

In order to set up these rules in a project we have to:
- import `.bazelrc`: https://github.com/swift-nav/gnss-converters-bazel/pull/3/files#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832f
- define `clang_format_config` target with `.clang-format` file: https://github.com/swift-nav/gnss-converters-bazel/pull/3/files#diff-7fc57714ef13c3325ce2a1130202edced92fcccc0c6db34a72f7b57f60d552a3

PR tested here: https://github.com/swift-nav/gnss-converters-bazel/pull/3

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,2 @@
+build:clang-format-check --aspects //bazel/clang_format:clang_format_check.bzl%clang_format_check_aspect
+build:clang-format-check --output_groups=report
--- a/clang_format/BUILD
+++ b/clang_format/BUILD
@@ -1,0 +1,27 @@
+load(":choose_clang_format.bzl", "choose_clang_format")
+
+choose_clang_format(
+    name = "clang_format_bin",
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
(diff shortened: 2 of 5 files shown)


isaactorz:
Awesome, a lot of this stuff is new to me too so I'll need a little bit of time to read up on aspects and go over this thoroughly.

PR #3, opened 2022-09-29
Initial swift_cc* wrappers

isaactorz:
Adds equivalents to our `swift_add*` wrappers in cmake.

The only one that's currently implemented is `swift_cc_test`:

The wrapper enforces that tests must be declared either `UNIT` or `INTEGRATION` so that they can be run separately.

In our cmake build this results in different targets: `do-all-unit-tests` and `do-all-integration-tests`.

The bazel way to achieve this is using `tag filters`:
```sh
bazel test --test_tag_filters=unit //...
```

Also note that with bazel we don't need any of the logic required in our cmake build to group the test targets together.

If you want to run all tests defined in the `WORKSPACE` (project): `bazel test //...`.

If you want to run all tests in a "package" (defined by a `BUILD.bazel` file): `bazel test //c/observation_adjuster:all`

If you want a single test target: `bazel test //c:gnss_converters:gnss_converters_test`.

The new wrappers are tested here: https://github.com/swift-nav/gnss-converters-private/pull/1340

### DIFF
--- a/swift_cc_defs.bzl
+++ b/swift_cc_defs.bzl
@@ -1,24 +1,24 @@
 def swift_cc_library(**kwargs):
-  """Wraps cc_library to enforce standards for a production library.
-  """
+    """Wraps cc_library to enforce standards for a production library.
+    """
     native.cc_library(**kwargs)
 
 def swift_cc_tool_library(**kwargs):
-  """Wraps cc_library to enforce standards for a non-production library.
-  """
+    """Wraps cc_library to enforce standards for a non-production library.
+    """
     native.cc_library(**kwargs)
 
 def swift_cc_binary(**kwargs):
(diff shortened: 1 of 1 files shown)


swift_cc_defs.bzl
```diff
@@ -0,0 +1,43 @@
+# Name for a unit test
+UNIT = "unit"
```
isaactorz, line 2: Bazel has a built in convention for splitting up tests that's similiar to how we distinguish unit vs integration tests.

The convention is `small` vs `large` tests: https://bazel.build/reference/test-encyclopedia#tag-conventions.

The you filter with `bazel test --test_size_filters=small`.

Putting this behind a constant means we can change this to the conventional way if we want without a big refactor.

krisukox (approved):

PR #5, opened 2022-09-30
Add Rules For Clang-Format [BUILD-305]

krisukox:
This PR adds rules for clang-format.

`clang-format-check` is implemented using the actions API - we can use cache functionality.
In order to check formatting in a repo use: `bazel build //... --config clang-format-check`
In order to format a repo use: `bazel run //bazel/clang_format`

explanation:
`clang-format-check` is defined in the `.bazelrc` file. Bazel passes all targets (`//...`) to the `clang_format/clang_format_check.bzl:clang_format_check_aspect` aspect. Bazel then iterates over files in the target and runs `clang-format` on it.
The clang format version is searched in the `clang_format/choose_clang_format.bzl` file. The clang format version is searched only once, and then we use it from the cache.

In order to set up these rules in a project we have to:
- import `.bazelrc`: https://github.com/swift-nav/gnss-converters-bazel/pull/3/files#diff-544556920c45b42cbfe40159b082ce8af6bd929e492d076769226265f215832f
- define `clang_format_config` target with `.clang-format` file: https://github.com/swift-nav/gnss-converters-bazel/pull/3/files#diff-7fc57714ef13c3325ce2a1130202edced92fcccc0c6db34a72f7b57f60d552a3

PR tested here: https://github.com/swift-nav/gnss-converters-bazel/pull/3

Original PR: https://github.com/swift-nav/swiftnav-bazel/pull/2

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,2 +1,0 @@
-build:clang-format-check --aspects //bazel/clang_format:clang_format_check.bzl%clang_format_check_aspect
-build:clang-format-check --output_groups=report
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,0 +1,2 @@
+build:clang-format-check --aspects //bazel/clang_format:clang_format_check.bzl%clang_format_check_aspect
+build:clang-format-check --output_groups=report
--- a/clang_format/run_clang_format.sh
+++ b/clang_format/run_clang_format.sh
@@ -12,7 +12,9 @@ format_all() {
         echo ".clang-format file not found. Bazel will copy the default .clang-format file."
         cp $CLANG_FORMAT_CONFIG .
     fi
-    git ls-files '*.[ch]' '*.cpp' '*.cc' '*.hpp' | xargs $CLANG_FORMAT_BIN -i
+    git describe --tags --abbrev=0 --always \
+    | xargs -I % git diff --diff-filter=ACMRTUXB --name-only --line-prefix=`git rev-parse --show-toplevel`/ % -- '*.[ch]' '*.cpp' '*.cc' '*.hpp' \
+    | xargs clang-format-14 -i
 }
 
 check_file() {
--- a/clang_format/run_clang_format.sh
+++ b/clang_format/run_clang_format.sh
@@ -1,11 +1,17 @@
 #! /bin/bash
 # Usages:
-# run_clang_format format_all <CLANG_FORMAT_BIN>
+# run_clang_format format_all <CLANG_FORMAT_BIN> <CLANG_FORMAT_CONFIG>
 # run_clang_format check_file <CLANG_FORMAT_BIN> <INPUT> <OUTPUT>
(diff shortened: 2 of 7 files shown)


krisukox:
So now, when we want to check formatting with the "default" `.clang-format` config file we have to run:
```
bazel build //... \
	--aspects @rules_swiftnav//clang_format:clang_format_check.bzl%clang_format_check_aspect \
	--output_groups=report
```
Custom  `.clang-format` file can be passed using the `--@rules_swiftnav//clang_format:clang_format_config` flag:
```
bazel build //... \
	--aspects @rules_swiftnav//clang_format:clang_format_check.bzl%clang_format_check_aspect \
	--output_groups=report \
	--@rules_swiftnav//clang_format:clang_format_config=//c:clang_format_config
```
I prepared a test PR with this: https://github.com/swift-nav/gnss-converters-private/pull/1373

If we want to format entire repo we can use `bazel run @rules_swiftnav//clang_format`. In the recent commit, I added a statement that checks if `.clang-format` exists in a repo. If it doesn't exist, bazel will copy the default `.clang-format` file to a repo.

I tried to pass `.clang-format` file with the `-style=file:<format_file_path>` flag (this feature has been added in [clang-format version 14](https://releases.llvm.org/14.0.0/tools/clang/docs/ClangFormat.html)), but I couldn't make it work.

clang_format/run_clang_format.sh
```diff
@@ -0,0 +1,33 @@
...
+
+format_all() {
+    CLANG_FORMAT_CONFIG=$(realpath $1)
+
+    cd $BUILD_WORKSPACE_DIRECTORY
+    if ! test -f .clang-format; then
+        echo ".clang-format file not found. Bazel will copy the default .clang-format file."
+        cp $CLANG_FORMAT_CONFIG .
+    fi
+    git ls-files '*.[ch]' '*.cpp' '*.cc' '*.hpp' | xargs $CLANG_FORMAT_BIN -i
```
isaactorz, line 15: One last thing before we merge this. This command is a little sluggish and will probably not perform that well in larger projects.

Can we implement is using `diff` like we do for the `clang-format-diff` target in cmake? :https://github.com/swift-nav/cmake/blob/2adba8029f2a1a476b963a32cd494cd03142b6df/ClangFormat.cmake#L238

.bazelrc
```diff
@@ -0,0 +1,2 @@
+build:clang-format-check --aspects //bazel/clang_format:clang_format_check.bzl%clang_format_check_aspect
+build:clang-format-check --output_groups=report
```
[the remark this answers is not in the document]
krisukox: Done

clang_format/run_clang_format.sh
krisukox: Done

isaactorz (approved):
LGTM

PR #7, opened 2022-10-13
Add qac compile commands script [BUILD-316]

krisukox:
This script changes the `-I` flag into the `-isystem` flag in `compile_commands.json`. As mentioned [here](https://swift-nav.atlassian.net/wiki/spaces/ENG/pages/2138934679/QAC+Helix+rules+analysis+and+approval?focusedCommentId=2141684606), Helix QAC needs the `-I` flag instead of `-isystem`.
The script can be run with: `bazel run //bazel:qac_compile_commands`
PR tested here: https://github.com/swift-nav/gnss-converters-bazel/pull/4

### DIFF
--- a/qac_compile_commands.sh
+++ b/qac_compile_commands.sh
@@ -3,7 +3,7 @@
 # This is needed for Helix QAC to work properly.
 # Usage: qac_compile_commands <REFRESH_COMPILE_COMMANDS>
 
-REFRESH_COMPILE_COMMANDS=$(realpath $1)
+REFRESH_COMPILE_COMMANDS=$1
 
 $REFRESH_COMPILE_COMMANDS
 sed -i 's/-isystem/-I/' $BUILD_WORKSPACE_DIRECTORY/compile_commands.json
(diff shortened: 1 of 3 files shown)


isaactorz (commented):
I think it would be nicer if we wrote a custom rule that expects to take a `compile_commands.json` file as input.

Then we could make it depend on the hedron rule, and we don't have to invoke two seperate commands to get what we're looking for.

krisukox:
I think we cannot create a rule that depends on the hedron rule because we have to run refresh_compile_commands, not only build.
But we can pass the hedron rule as the data parameter so we can invoke it from our script.

qac_compile_commands.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
+# This script turns the -isystem flag into the -I flag.
+# This is needed for Helix QAC to work properly.
+# Usage: qac_compile_commands <REFRESH_COMPILE_COMMANDS>
+
+REFRESH_COMPILE_COMMANDS=$(realpath $1)
```
isaactorz, line 6: I don't think using realpath is necessary here.

Since we make the compile_commands target a data dependency, the python script should be copied into the runtime environment.

(I tested this locally and it seemed to work).

krisukox: I needed it in the previous version of this script but, now it is not necessary.

isaactorz (approved):

PR #12, opened 2022-11-09
Improve configure_file and add rapidcheck, json, yaml-cpp [BUILD-411]

krisukox:
This PR improves behavior of the configure_file function.

In the improvement I added `#cmakedefine01` which is used e.g. [here](https://github.com/swift-nav/starling-core/blob/master/starling_core_build_config/config.h.in#L18)
I also fixed the issue with identical substring as one of the string e.g.:
```
#cmakedefine01 ABC
#cmakedefine01 ABC_STRING
```
In the example above, `ABC` would be changed in both cases. That's why I added a newline `\n` in one of the matches.
I also had to add a match with a space in order to cover this case:
```
#cmakedefine ABC @ABC@
```
PR also adds support for the following third_party library:
- rapidcheck
- nlohmann/json
- yaml-cpp

PR tested here:
https://github.com/swift-nav/starling/pull/7124
https://github.com/swift-nav/starling-core/pull/100


third_party/rapidcheck.BUILD
```diff
@@ -0,0 +1,23 @@
...
+    default_visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]),
+)
+
+cmake(
+    name = "rapidcheck",
```
isaactorz, line 13: I looked at the build. Doesn't seem  especially complicated. Any reason not to use pure bazel?

I prefer we use bazel except when it would be obnoxious (like check).

krisukox: I thought it will be hard to add it but indeed it was straightforward.

krisukox:
I extracted `configure_file` to another function in order to define rule in as easier way.
https://github.com/swift-nav/starling/pull/7124/files#diff-6eb3c92823224db3748c0f7201f8c0495d8f736d806eb304f14273981b2d7ccb

isaactorz (approved):
LGTM!

PR #18, opened 2022-11-24
Add macro for setting up local include prefix

isaactorz:
Defines a macro for setting up the local include prefix so that we don't have to reimplement this logic in every repository.

I've tested it with `libsbp` (as a standalone project), and with `gnss-converters` (using `libsbp` as an external repo), and it works as expected.

I'm also working on integrating this into `swift_cc*` so that these paths get set up automatically but I still need to figure out how to deal with `$(GENDIR)`. See: https://github.com/swift-nav/rules_swiftnav/pull/19.

In the mean time this will do.


krisukox (approved):
I've tested it locally, and it works well.

isaactorz:
Closing in favor of: https://github.com/swift-nav/rules_swiftnav/pull/19

PR #19, opened 2022-11-24
cc: Add local_includes attribute to swift_cc*

isaactorz:
# Description

Adds a `local_includes` attribute that automatically handles constructing the correct include paths for both local and external uses.

When a bazel workspace is consumed externally, (i.e. via `local_repository`), its sources are placed under <execroot>/external/<workspace_root>/. This typically breaks local include paths defined using -I. 

To help us deal with this I've added a new macro `construct_local_include` in the new `cc//:utils.bzl` that handles this "nuance". Thanks to @krisukox for originally showing me this workaround. 

It also deals with `$(GENDIR)` which requires slightly different path construction. The assumption is that files will be generated into the root of `$(GENDIR)` for this to work.

Using this macro is a bit awkward and it's likely that users won't remember to add it so I've introduced a new attribute to our `swift_cc` wrappers: `local_includes`, that handles setting this up by default. Moving forward we should prefer using this attribute to passing include paths using `copts`.

As an aside it's occurred to me that these custom manipulations we do for `copts` probably will not play nicely if we need to `select` over them. Something to keep in mind.

# Testing

Tested here with libsbp:
- https://github.com/swift-nav/libsbp/tree/itorres/external-include-prefix


silverjam (approved):
lgtm

PR #20, opened 2022-11-28
Try self-hosted runer

isaactorz:
Test out self-hosted runner.


isaactorz:
Probably needs a lot more effort to get this going. Closing for now.

PR #25, opened 2022-12-13
Add common compilation flags to tests [BUILD-448]

krisukox:
This PR adds common compilation flags to `swift_cc_test_library` and adds `common_c_opts` flag to `swift_cc_test` which adds common compilation flags.

This PR sets the `linkstatic` flag to `True` in `swift_cc_test`.

Tested here:
- https://github.com/swift-nav/libsbp/pull/1269
- https://github.com/swift-nav/gnss-converters-private/pull/1407
- https://github.com/swift-nav/libswiftnav-private/pull/413
- https://github.com/swift-nav/libpal/pull/286

### DIFF
--- a/cc/defs.bzl
+++ b/cc/defs.bzl
@@ -202,7 +202,7 @@ def swift_cc_test_library(**kwargs):
     """Wraps cc_library to enforce Swift test library conventions.
 
     Args:
-        **kwargs: See https://bazel.build/reference/be/c-cpp#cc_test
+        **kwargs: See https://bazel.build/reference/be/c-cpp#cc_library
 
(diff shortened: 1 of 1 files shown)


jungleraptor (dismissed):
On second thought lets go ahead and merge this.

But I think our goal once we get everything building with these compile flags should be to remove the bool flag.

jungleraptor (commented):
I went back and forth but lets discuss a little more in our meeting tomorrow. Basically I don't want to have that bool flag become a permanent part of the API and am a little concerned that it might become difficult to remove it if we merge these changes as is.

PR #26, opened 2022-12-15
third_party: Add build file for openssl [BUILD-376]

isaactorz:
Adds a build file for building openssl with bazel.

Tested on x64 mac and linux.


silverjam (approved):
🤘

PR #28, opened 2022-12-20
cc: selectable common copts [BUILD-440]

isaactorz:
Makes the common compile options we build with `select`-able, so that we can enable building with a variety of toolchains.

This enables us to build with, for example, gcc-6 which doesn't support the `-Wimplicit-fallthrough` flag.

# Usage

Requires defining a `platform` that "activates" the desired `constraint_value`:

```bazel
platform(
    name = "gcc-6",
    constraint_values = [
        "@rules_swiftnav//cc/constraints:gcc-6",
        "@platforms//os:linux",
        "@platforms//cpu:x86_64"
    ],
)
```

To build:
`CC=gcc-6 CXX=g++6 bazel build --platforms=//:gcc-6 //...`

# Design Notes

In our cmake build the list of compiler flags is implicit and dynamic. For every option we check that the compiler can build a simple program with that option enabled before adding it to the compile line. Here the compile flags are an explicit static list of options.

Technically it would also be possible to do this in Bazel, but it violates the design principles of keeping the build hermetic, which is a big advantage of using bazel in the first place. Also we would no longer have been able to use the default c++ toolchain.

Note that this assumes that compiler versions we build with is ok to be public since this is being added to our public rules repo.


cc/copts.bzl
```diff
@@ -0,0 +1,50 @@
...
+    "-Wunused-value",
+    "-Wunused-variable",
+    "-Wvolatile-register-var",
+    "-Wwrite-strings",
+    "-Wno-error=deprecated-declarations",
+    #TODO: [BUILD-405] - Figure out why build breaks with this flag
+    #"-Wmissing-include-dirs"
+]
+
+GCC6_COPTS = [copt for copt in DEFAULT_COPTS if copt != "-Wimplicit-fallthrough"]
```
silverjam, line 50: Nit, maybe consider a more expandable form like:

```suggestion
GCC6_DISABLED_COPTS = ["-Wimplicit-fallthrough"]
GCC6_COPTS = [copt for copt in DEFAULT_COPTS if copt in GCC6_DISABLED_COPTS]
```

cc/constraints/BUILD.bazel
```diff
@@ -0,0 +1,8 @@
+package(default_visibility = ["//visibility:public"])
+
+constraint_setting(name = "compiler")
+
+constraint_value(
+    name = "gcc-6",
+    constraint_setting = ":compiler",
```
silverjam, line 7: Does this override a standard setting that Bazel uses to select a compiler?  Is specifying `gcc-6` enough to allow Bazel to locate all associated tools like `g++-6`?

silverjam: I guess just pointing me at the docs where that explains this would work too :-D

silverjam (approved):

cc/constraints/BUILD.bazel
isaactorz: It doesn't override any standard Bazel settings. Out of the box this attribute alone won't have any affect on the toolchain that is selected.

You need to pair it with a custom `platform` target and `toolchain` target for that.

In the example above I provided the minimal amount of code necessary to trigger this condition so that it "activates" this constraint.

In Bazel a `platform` basically defines a bundle of `constraint_values`. When the platform is activated via `--platform`, those constraint values are active.

Bazel selects the first toolchain that it knows about that matches all of the active `constraint_values`. In the example above the platform I defined is the minimum set of constrain values to "match" the default cc toolchain, plus our own custom constraint setting that we defined here. I still had to specify the gcc-6 toolchain using `CC` and `CXX`.

So how I imagine this being used is we might have a custom `debian_stretch` platform that defines some constraint settings, and a corresponding `gcc-6` toolchain that is selected when that platform is activated.

We could also end up supporting another target platform, `foo`, that corresponds to a yocto toolchain or something that also uses gcc-6.

In both cases the build system will still choose the right set of flags for the compile options.

isaactorz: Here's the official documentation on `constraints` and `platforms`: https://bazel.build/extending/platforms#constraints-platforms

silverjam: Cool, thanks

wwieclaw (approved):

PR #29, opened 2022-12-23
Added build rules for 4 third party libs

wwieclaw:
* zlib
* hdf5
* netcdf-c
* netcdf-cxx


third_party/hdf5.BUILD
```diff
@@ -0,0 +1,363 @@
+'''
+hdf5@1.12.2
+'''
+
```
isaactorz, line 4: Let's add a comment pointing to where we got this from (I'm assuming it was tensorflow?). See here for an example: https://github.com/swift-nav/rules_swiftnav/blob/cb09e9ada2a509827143db3f192ca1f0bac30b9c/third_party/openssl.BUILD#L11.

third_party/hdf5.BUILD
```diff
@@ -0,0 +1,363 @@
...
+        "#define H5_SIZEOF_UINT_FAST8_T 1",
+        "#define H5_SIZEOF_UINT_LEAST16_T 2",
+        "#define H5_SIZEOF_UINT_LEAST32_T 4",
+        "#define H5_SIZEOF_UINT_LEAST64_T 8",
+        "#define H5_SIZEOF_UINT_LEAST8_T 1",
+        "#define H5_SIZEOF_UNSIGNED 4",
+        "#define H5_SIZEOF__QUAD 0",
+        "#define H5_STDC_HEADERS 1",
+        "#define H5_USE_110_API_DEFAULT 1",
+        '#define H5_VERSION "1.10.6"',
```
isaactorz, line 322: This seems to not match what we have above.

third_party/hdf5.BUILD
```diff
@@ -0,0 +1,363 @@
...
+    deps = [
+        "@zlib",
+    ],
+)
+
+genrule(
+    name = "H5pubconf_h",
+    outs = ["config/H5pubconf.h"],
+    cmd = "\n".join([
+        "cat <<'EOF' >$@",
```
isaactorz, line 81: These settings all look pretty reasonable.

I think it'll probably be sufficient for all the platforms we're targeting for this library (x64,aarch64 variants of macos and linux).

third_party/hdf5.BUILD
```diff
@@ -0,0 +1,363 @@
...
+        "char H5libhdf5_settings[]=\"swiftnav\";",
+        "EOF",
+    ]),
+)
+
+genrule(
+    name = "H5Tinit_c",
+    outs = ["config/H5Tinit.c"],
+    cmd = "$(location :H5detect) $@",
```
isaactorz, line 344: What is the content of the file this binary generates?

third_party/netcdf-c.BUILD
```diff
@@ -0,0 +1,31 @@
...
+        "BUILD_SHARED_LIBS": "ON",
+        "CMAKE_BUILD_TYPE": "Release",
+        "CMAKE_INSTALL_LIBDIR": "lib",
+        "ENABLE_DAP": "OFF",
+        "ENABLE_EXAMPLES": "OFF",
+        "ENABLE_PNETCDF": "OFF",
+        "ENABLE_TESTS": "OFF",
+        "USE_SZIP": "OFF",
+    },
+    generate_args = ["-GNinja"],
```
isaactorz, line 24: Let's stick with make for all of our `cmake` dependencies.

Ninja is great for incremental rebuilds, but for a fresh build there's not much advantage over pure make. Since these are third party dependencies and won't change often, I think we can live without. One less dependency for our build to have to concern itself with.

third_party/netcdf-c.BUILD
```diff
@@ -0,0 +1,31 @@
...
+    lib_source = ":allsrcs",
+    out_shared_libs = ["libnetcdf.so"],
```
isaactorz, line 26: We should use static linking for all of our third-party dependencies so we don't have to worry about providing these dependencies in our deployments.

third_party/netcdf-cxx.BUILD
```diff
@@ -0,0 +1,17 @@
...
+
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]),
+)
+
+cmake(
+    name = "netcdf-cxx",
+    lib_source = ":srcs",
+    out_shared_libs = ["libnetcdf-cxx4.so.1.1.0"],
```
isaactorz, line 15: static linking

third_party/netcdf-cxx.BUILD
```diff
@@ -0,0 +1,17 @@
...
+
+load("@rules_foreign_cc//foreign_cc:defs.bzl", "cmake")
+
+filegroup(
+    name = "srcs",
+    srcs = glob(["**"]),
+)
+
+cmake(
+    name = "netcdf-cxx",
```
isaactorz, line 13: Do we need to provide `deps` here for the c-lib? Is it still a WIP?

third_party/hdf5.BUILD
isaactorz: And our copyright notice.

isaactorz (commented):
Comments below.

third_party/netcdf-cxx.BUILD
wwieclaw: Yeah my bad. Idk how but a everything builds just fine even though I didn't specify the deps. I'll try to look into it.

third_party/hdf5.BUILD
wwieclaw: https://pastebin.com/rjzdCntD

It has some machine/date specific things:
```
 * Created:		Jan  2, 2023
 *			Wiktor Więcław <wiktor@wiktor-Latitude-5521>
```

wwieclaw: I guess we could hardcode the source file if we find this to be a problem, right?

isaactorz: Maybe we can do it in a separate PR but we will probably want to revisit this if causes problems with caching (the timestamps and user info will probably invalidate the cache).

third_party/netcdf-c.BUILD
```diff
@@ -0,0 +1,41 @@
...
+filegroup(
+    name = "allsrcs",
+    srcs = glob(["**"]),
+)
+
+cmake(
+    name = "netcdf-c",
+    cache_entries = {
+        "BUILD_SHARED_LIBS": "OFF",
+        "CMAKE_BUILD_TYPE": "Release",
```
isaactorz, line 26: Did you need this to build? We might want to check if `rules_foriegn_cc` handles the build type for us (i.e. if we use `bazel build -c dbg` it will automatically pass the corresponding option `-DCMAKE_BUILT_TYPE=Debug` to cmake.

isaactorz (commented):
Couple more small comments

third_party/netcdf-c.BUILD
wwieclaw: nope, looks like it builds fine without it

third_party/hdf5.BUILD
wwieclaw: https://swift-nav.atlassian.net/browse/BUILD-499

isaactorz (approved):

PR #32, opened 2023-01-18
Add exceptions parameter to swift_cc* [BUILD-498]

krisukox:
This PR:
- disables exceptions by default in all `swift_cc*` rules except `swift_cc_test`,
- adds exceptions parameter to all `swift_cc*` rules except `swift_cc_test`,
- adds common compilation flags to `swift_cc_test_library`.


isaactorz:
@krisukox I think we can close this

PR #33, opened 2023-01-18
Added -ldl to hdf5 [BUILD-511]

wwieclaw:
See https://swift-nav.atlassian.net/jira/software/projects/BUILD/boards/227?selectedIssue=BUILD-511

Tested on orion-engine's CI pipeline

### DIFF
--- a/third_party/hdf5.BUILD
+++ b/third_party/hdf5.BUILD
@@ -45,7 +45,7 @@ cc_library(
         "src",
         "src/H5FDsubfiling",
     ],
-    linkopts = [],
+    linkopts = ["-ldl"],
     visibility = ["//visibility:public"],
     deps = [
         ":header",


third_party/hdf5.BUILD
```diff
@@ -45,7 +45,7 @@ cc_library(
         "src",
         "src/H5FDsubfiling",
     ],
-    linkopts = [],
+    linkopts = ["-ldl"],
```
isaactorz, line 48: Let's remove this from here and add it to the target that actually needs it.

Also this is likely not cross-platform (will probably fail on macos).

wwieclaw: > Let's remove this from here and add it to the target that actually needs it.

Why? It makes it a bit harder on the users of the library, do you just want to save some link time?

> Also this is likely not cross-platform (will probably fail on macos).

According to https://stackoverflow.com/questions/20169660/where-is-libdl-so-on-mac-os-x there is no need to link any library to use `<dlfcn.h>`, so I guess we could just use `select`, right? Idk about windows.

isaactorz:
@wwieclaw I think we can close this one right?

PR #34, opened 2023-01-20
Filter out .inc extension

krisukox:
This PR allows filtering out `.inc` extension in the Clang Tidy aspect.


wwieclaw (approved):

isaactorz (approved):

isaactorz:
@krisukox I took a quick scan through the code and it doesn't look like we're doing this already, but might also want to consider filtering out protobuf generated files (`.pb.cc`).

I think we should avoid linting generated files in general. Maybe we can use a similiar technique as your other PR (we can filter out files that are located in the `$(GENDIR)`).

PR #35, opened 2023-01-20
Fix include paths in Clang Tidy rule

krisukox:
The problem:
Bazel adds `-isystem` to all include paths, which leads to inaccurate Clang Tidy results. (Clang Tidy doesn't check our headers)

The solution:
We will distinguish system paths by checking if the path starts with `external/` or contains `/bin/external/`.

Testing:
I ran it in Starling and tested if the headers are linted.


isaactorz (approved):
Nice.

PR #36, opened 2023-01-23
Fix the default use of clang-format config file

krisukox:
This PR fixes the problem with the default clang-format config file.


isaactorz (approved):
🙏🏼

PR #37, opened 2023-01-23
clang-tidy-rule: Filter out headers

krisukox:
[Here in gnss-converters](https://github.com/swift-nav/gnss-converters-private/blob/master/c/gnss_converters/BUILD.bazel) we pass headers as source files. In `clang-tidy-rule` we iterate over sources and run clang-tidy on them.

Since a header is not a final compilation unit, we cannot be sure if this will work. If we want to keep passing headers as source files, we have to filter them out in `clang-tidy-rule`.

### DIFF
--- a/clang_tidy/clang_tidy.bzl
+++ b/clang_tidy/clang_tidy.bzl
@@ -143,7 +143,7 @@ def _clang_tidy_aspect_impl(target, ctx):
     compilation_contexts = _get_compilation_contexts(target, ctx)
 
     srcs = _rule_sources(ctx)
-    unsupported_ext = ["inc"]
+    unsupported_ext = ["inc", "h", "hpp"]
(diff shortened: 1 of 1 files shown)


wwieclaw (approved):

isaactorz (commented):
The recommended approach in bazel is to put public headers in `hdrs` and private headers in `srcs`: https://bazel.build/reference/be/c-cpp#hdrs, so we should definitely support this.

I'm assuming these files will still end up linted so long as they're part of the include path of the `.c/.cc` files we pass to clang-tidy?

If that's the case I think we should leave a comment that explains the files will still be linted even if we exclude them here.

krisukox:
Yes, headers will be linted if included in a source file.
I added a comment.

PR #38, opened 2023-01-23
cc: move -Werror to a toolchain feature

isaactorz:
Bazel 6.0 introduces a new toolchain feature;
treat_warnings_as_errors which enables toggling -Werror on a per target basis, as well as at the command line.

### DIFF
--- a/cc/defs.bzl
+++ b/cc/defs.bzl
@@ -63,10 +63,11 @@ def swift_cc_library(**kwargs):
 
     copts = _common_c_opts(nocopts, pedantic = True)
     copts = local_includes + copts
-    kwargs["copts"] = copts + (kwargs["copts"] if "copts" in kwargs else [])
+    kwargs["copts"] = copts + kwargs.get("copts", [])
 
-    kwargs["features"] = _default_features() + (kwargs["features"] if "features" in kwargs else [])
-    kwargs["tags"] = (kwargs["tags"] if "tags" in kwargs else []) + [LIBRARY]
+    kwargs["features"] = _default_features() + kwargs.get("features", [])
+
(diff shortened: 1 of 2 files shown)


cc/defs.bzl
```diff
@@ -94,6 +99,8 @@ def swift_cc_tool_library(**kwargs):
     copts = local_includes + copts
     kwargs["copts"] = copts + (kwargs["copts"] if "copts" in kwargs else [])
 
+    kwargs["features"] = _default_features() + (kwargs["features"] if "features" in kwargs else [])
```
silverjam, line 102: Does Starlark support `dict.get`?

https://docs.python.org/3/library/stdtypes.html#dict.get

If it I think this would become:

```suggestion
    kwargs["features"] = _default_features() + kwargs.get("features", [])
```

isaactorz: It does indeed: https://bazel.build/rules/lib/dict#get.

Thanks this is much better.

silverjam (approved):

krisukox (approved):

cc/defs.bzl
```diff
@@ -34,6 +34,14 @@ def _common_c_opts(nocopts, pedantic = False):
 def _construct_local_includes(local_includes):
     return [construct_local_include(path) for path in local_includes]
 
+def _default_features():
+    return select({
+        # treat_warnings_as_errors passes the option -fatal-warnings
+        # to the linker which ld on mac does not understand.
+        "@platforms//os:macos": [],
```
isaactorz, line 41: I discovered this issue when trying to build my branch on mac.

Seems like it's something that will need to get patched upstream.

We're going to be using our own implementation shortly though so we will also be able to fix it ourselves to as to not have to wait on it to land.

PR #39, opened 2023-01-24
Add sonarcloud config rule [BUILD-498]

krisukox:
This PR adds:
- the `gen_sonar_cfg` rule that generates sonarcloud config,
- the `_get_cov_files` aspect, which returns files of a target that should be analyzed by sonarcloud,
- `filegroup` inside `swift_cc_test` that contains tests source files. 

`_get_cov_files`:
- checks whether the target is created with `swift_cc_library` or `swift_cc_test_library`,
- filters out generated files.

`gen_sonar_cfg` takes 3 arguments:
- `targets` - all libraries that should be analyzed by sonarcloud,
- `test_srcs` - all test sources,
- `root_dir` - `string_flag` that is used to properly generate a config file.

**Tested here:** https://github.com/swift-nav/starling-core/pull/185

**Example usage:**
```
BUILD.bazel:
gen_sonar_cfg(
    name = "gen_sonar_cfg",
    targets = ["//:lib"],
    test_srcs = ["//:lib_test_srcs"],
)
```
`bazel build //:gen_sonar_cfg --@rules_swiftnav//tools:root_dir=$PWD`


tools/get_coverage_files.bzl
```diff
@@ -0,0 +1,82 @@
...
+)
+
+def _get_cov_files_impl(target, ctx):
+    files = []
+
+    if not CcInfo in target:
+        return [FilesInfo(files = [])]
+
+    tags = getattr(ctx.rule.attr, "tags", [])
+    if not LIBRARY in tags and not TEST_LIBRARY in tags:
```
isaactorz, line 17: Shouldn't we also grab files from targets that are tag `BINARY`?

tools/get_coverage_files.bzl
```diff
@@ -0,0 +1,82 @@
...
+        echo 'sonar.sources=\\' >> {out_file}
+        first=1
+        for file in {all_files}; do
+            if [[ $first -eq 1 ]]
+            then
+                first=0
+            else
+                echo ',\\' >> {out_file}
+            fi
+            echo -n "  {root_dir}/$file" >> {out_file}
```
isaactorz, line 63: So is what we're doing with the `root_dir` var making sure that the sources original absolute path?

cc/defs.bzl
```diff
@@ -215,6 +223,18 @@ def swift_cc_test(name, type, **kwargs):
 
     _ = kwargs.pop("nocopts", [])  # To handle API compatibility.
 
+    srcs_name = name + "_srcs"
+    srcs = kwargs.get("srcs", [])
+
+    native.filegroup(
```
isaactorz, line 229: I think it might be a little confusing to users or non-obvious that we generate an additional target for the sources.

Some docstring like I've provided in the `swift_cc` macros would go a long way here.

See this project for how to add documentation to rules: https://github.com/bazelbuild/stardoc

isaactorz (commented):
A couple of small questions.

One overall thing though, is to consider not using `coverage` and `cov` in the names. Coverage Uploads and the static analysis that uses the `sonar-scanner` are two different features.

Most of this stuff is related to the static analysis.

tools/BUILD.bazel
```diff
@@ -8,7 +8,15 @@
...
 
+load("@bazel_skylib//rules:common_settings.bzl", "string_flag")
+
 exports_files(
     glob(["*.bzl"]) + ["qac_compile_commands.sh"],
     visibility = ["//visibility:public"],
 )
+
+string_flag(
+    name = "root_dir",
```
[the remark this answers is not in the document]
krisukox: We can, but I think we might want to use it in other tools. We can leave it as it is and use it as a common parameter.

tools/get_coverage_files.bzl
krisukox: Yes, that's right.

tools/get_coverage_files.bzl
krisukox: Yes, we should. Sorry, I removed it when I was working on the new design.

cc/defs.bzl
krisukox: Added

tools/get_coverage_files.bzl
isaactorz: I think that might cause a problem for caching since the `sonar-project.proprties` will be different from machine to machine.

But we can probably live with that for now.

isaactorz (approved):
LGTM

PR #41, opened 2023-02-02
Change qac script and add get_target_files aspect [BUILD-506]

krisukox:
This PR changes the `qac_compile_commands` script so that it adds `-isystem` before external include paths and `-I` otherwise.
It also adds `get_target_hdrs` aspect.

### DIFF
--- a/tools/get_files.bzl
+++ b/tools/get_files.bzl
@@ -1,0 +1,57 @@
+load("//cc:defs.bzl", "BINARY", "LIBRARY", "TEST_LIBRARY")
+
+FilesInfo = provider(
+    fields = {
+        "files": "files of the target",
+    },
+)
+
+def _get_hdrs(ctx):
+    files = []
+
+    if hasattr(ctx.rule.attr, "hdrs"):
+        for hdr in ctx.rule.attr.hdrs:
+            for file in hdr.files.to_list():
+                if not file.path.startswith(ctx.genfiles_dir.path):
+                    files.append(file)
+    return files
+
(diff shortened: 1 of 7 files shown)


tools/get_files.bzl
```diff
@@ -0,0 +1,57 @@
...
+    if not LIBRARY in tags and not TEST_LIBRARY in tags and not BINARY in tags:
+        return [FilesInfo(files = [])]
+
+    files.extend(_get_srcs(ctx))
+
+    files.extend(_get_hdrs(ctx))
+
+    return [FilesInfo(files = files)]
+
+get_target_files = aspect(
```
isaactorz, line 45: I think we should make the names a little more C++ specific:
`get_cc_target_files`, `get_cc_target_hdrs`, `get_cc_files.bzl` etc...

tools/qac_compile_commands.py
```diff
@@ -0,0 +1,28 @@
+import sys, subprocess, json, os
```
isaactorz, line 1: Let's add a `__main__` function that intializes `refresh_compile_commands` and `workspace_dir`, and then passes those parameters to a `main()` function to tidy up this script.

Also make sure it's formatted with `black`.

Lastly add some doc on what this script in doing (similiar to the doc I added in the qac script).

isaactorz (commented):
Few nits but mostly looks good!

The `get_files` aspect is quite handy.

isaactorz (approved):
LGMT

PR #42, opened 2023-02-09
Add doxygen rule [BUILD-450]

krisukox:
This PR adds doxygen rule that uses system doxygen to generate documentation. It also checks if `dot` is installed in the system and sets path in the doxygen config file.

Tested here:
https://github.com/swift-nav/starling/pull/7536
https://github.com/swift-nav/starling-core/pull/213
https://github.com/swift-nav/libpal/pull/293


tools/doxygen.bzl
```diff
@@ -0,0 +1,70 @@
...
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+load("//tools:configure_file.bzl", "configure_file_impl")
+
+def _swift_doxygen_impl(ctx):
+    vars = ctx.attr.vars | {}
```
krisukox, line 14: This is needed in order to copy dict instead of referencing it. Without this I received:
`Error: trying to mutate a frozen dict value`

tools/doxygen.bzl
```diff
@@ -0,0 +1,73 @@
...
+    for src_dir in in_list:
+        if first_dir:
+            first_dir = False
+        else:
+            string += '" "'
+        string += src_dir
+    return string
+
+def _swift_doxygen_impl(ctx):
+    vars = ctx.attr.vars | {}
```
krisukox, line 26: This is needed in order to copy dict instead of referencing it. Without this, I received:
`Error: trying to mutate a frozen dict value`

tools/doxygen.bzl
```diff
@@ -0,0 +1,62 @@
...
+    implementation = _swift_doxygen_impl,
+    attrs = {
+        "vars": attr.string_dict(),
+        "template": attr.label(
+            allow_single_file = [".in"],
+            mandatory = True,
+        ),
+        "deps": attr.label_list(),
+        "doxygen_source_directories": attr.string_list(),
+        "doxygen_output_directory": attr.string(),
```
isaactorz, line 55: In your test PR's I see a couple of different conventions for specifying the output directory (some use `doxygen`, and some use `<lib>_doxygen`. Do you think it makes more sense to just standardize on something and make that the default value here instead?

I'm thinking what would be more intuitive for people who end up using this rule and then are trying to figure out where the documentation was generated to.

tools/doxygen.bzl
isaactorz: I'd add this as a comment in the code

tools/doxygen.bzl
isaactorz, line 1: `tools` will get pretty overloaded if we dump everything there. We should make a seperate `doxygen` folder.

tools/doxygen.bzl
krisukox: I removed this parameter. Output directory names are created from `target_name + "_doxygen"`.

isaactorz (dismissed):
~~LGTM!~~

Edit: not so fast.

tools/configure_file.bzl
```diff
@@ -10,7 +10,7 @@
 
 CMAKE_FALSE_CONSTANTS = ["0", "OFF", "NO", "FALSE", "N", "IGNORE", "NOTFOUND"]
 
-def configure_file_impl(ctx, vars):
+def configure_file_impl(ctx, vars, out):
```
isaactorz, line 13: Can we do this without breaking the API?

It's going to be a pain to get these changes in as we'll have to merge this update from the top down before we can merge the PR's that implement this.

I think this can be avoided if we just add a `out` attr to the doxygen rule so this can continue to be referenced as `ctx.attr.out`

krisukox: But then we would have to pass the `out` parameter to the `swift_doxygen` rule (because we can't overwrite an attribute inside a rule), which I think might complicate the doxygen rule API. (The `out` parameter is used to generate the Doxyfile "template", not the doxygen output directory)

How about introducing None as a default parameter and if any parameter is not set, taking it from the ctx:
```
def configure_file_impl(ctx, vars = None, out = None):
    if vars == None:
        vars = ctx.attr.vars

    if out == None:
        out = ctx.attr.out
```

isaactorz: Yeah that works 👍🏼

isaactorz (approved):
lgtm!

PR #43, opened 2023-02-09
cc: toolchain: hermetic llvm toolchain [BUILD-519]

isaactorz:
Adds a hermetic llvm toolchain for x86_64-linux.

# Usage

In your `WORKSPACE.bazel` file:
```bazel
load("@rules_swiftnav//cc:repositories.bzl", "swift_cc_toolchain", "register_swift_cc_toolchains")

# create rule for downloading the llvm distribution
swift_cc_toolchain()

# register the toolchain
register_swift_cc_toolchains()
```

# Design Notes

The original idea was to vendor https://github.com/grailbio/bazel-toolchain,
but this would've pulled in thousands of lines of complicated toolchain config code we probably don't need for our builds. It's an open source project that strives to be highly configurable for different use cases, and support a wide variety of llvm toolchains.

So instead I basically wrote this from the ground up, and leaned heavily on ideas from the above and similar projects.

The result is a much simpler and maintainable setup that is completely understood what every piece is for. Notably, it doesn't rely on any repository rules, template expansion, or symlink shenanigans.

Currently only x64-linux is supported but this is most of the general framework
we'll need to expand to supporting mac, and potentially even some of our cross-compilation toolchains.

# Testing

- [x] Test intel mac build
- [x] Test arm mac build
- [x] Test release configuration build
- [x] Test debug configuration build
- [x] Test coverage build
- [x] Test remote caching
- [x] Test whether `-fdebug-prefix` is necessary

Test PR's:
- https://github.com/swift-nav/starling/pull/7608
- https://github.com/swift-nav/orion-engine/pull/5545

# References

Various resources that helped me put this all together:
- https://ltekieli.com/cross-compiling-with-bazel/
- https://github.com/grailbio/bazel-toolchain
- https://github.com/bazelembedded/bazel-embedded
- https://github.com/bazelbuild/bazel/blob/master/tools/cpp/unix_cc_configure.bzl


cc/repositories.bzl
```diff
@@ -0,0 +1,26 @@
...
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+#
+# THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+# EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+# WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+LLVM_DISTRIBUTION_URL = "https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.0/clang%2Bllvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04.tar.xz"
```
isaactorz, line 13: This will change once we can get our own llvm builds up on GH releases.

cc/repositories.bzl
```diff
@@ -0,0 +1,26 @@
...
+
+def swift_cc_toolchain():
+    if "llvm-distribution" not in native.existing_rules():
```
notoriaga, line 16: I've seen other things like this use [maybe](https://bazel.build/rules/lib/repo/utils#maybe) for doing something something similar, or would that only catch an exisiting llvm distribution with the exact same version?

isaactorz: No your right maybe would serve the same purpose hear and is probably a little more idiomatic! Fixed

cc/toolchain/cc_toolchain_config.bzl
```diff
@@ -0,0 +1,171 @@
...
+        "-Wthread-safety",
+        "-Wself-assign",
+    ]
+
+    # -fstandalone-debug disables options that optimize
+    # the size of the debug info.
+    # https://clang.llvm.org/docs/UsersManual.html#cmdoption-fstandalone-debug
+    dbg_compile_flags = ["-g", "-fstandalone-debug"]
+
+    opt_compile_flags = [
```
joelynch, line 69: Great that all this complexity is hidden away behind a function with a really simple interface

cc/toolchain/cc_toolchain_config.bzl
```diff
@@ -0,0 +1,171 @@
...
+        # Will need to validate with some testing.
+        #"-fdebug-prefix-map={}=__bazel_toolchain_llvm_repo__/".format(toolchain_path_prefix),
+    ]
+
+    # Coverage flags:
+    coverage_compile_flags = ["-fprofile-instr-generate", "-fcoverage-mapping"]
+    coverage_link_flags = ["-fprofile-instr-generate"]
+
+    # true if using lld
+    supports_start_end_lib = True
```
joelynch, line 144: some of these might be able to be just passed as keyword args rather than defining a variable first but this is personal preference

cc/toolchain/BUILD.bazel
```diff
@@ -0,0 +1,121 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+load(":cc_toolchain_config.bzl", "cc_toolchain_config")
+
+filegroup(name = "empty")
+
+filegroup(
+    name = "wrappers",
+    srcs = glob([
```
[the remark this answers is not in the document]
isaactorz: Yeah there's [buildifier](https://github.com/bazelbuild/buildtools), in fact the CI check just makes sure all files are formatted with it.

cc/toolchain/BUILD.llvm
```diff
@@ -0,0 +1,132 @@
+# Copyright (C) 2022 Swift Navigation Inc.
```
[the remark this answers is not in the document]
isaactorz: Yeah good idea

cc/toolchains/BUILD.bazel
```diff
@@ -0,0 +1,121 @@
...
+    srcs = [
+        ":wrappers",
+        "@llvm-distribution//:ar",
+    ],
+)
+
+filegroup(
+    name = "as_files",
+    srcs = [
+        ":wrapppers",
```
krisukox, line 35: Typo: `ppp`

krisukox (approved):
LGTM

cc/toolchains/BUILD.bazel
isaactorz: Good catch!

PR #44, opened 2023-02-21
Added bazel native build of netcdf-c and netcdf-cxx [BUILD-530]

wwieclaw:
# Design Notes
There are 3 header files generated by CMake build and each of those file has a corresponding target:
* `include/netcdf_dispatch.h.in` - `expand_netcdf_dispatch_h_in`
* `include/netcdf_meta.h.in` - `expand_netcdf_meta_h_in`
* `config.h.in`- `gen_config_h`

The first two templates I expand using `//tools:configure_file` with the same values from the file that CMake build generated in the docker image defined [here](https://github.com/swift-nav/netcdf-c-bazel/blob/main/Dockerfile). The third file is entirely copied from CMake build.

There are 3 `.m4` files in `libsrc` that transformed to source fiiles. For this I used a simple genrule:
```
genrule(
    name = "gen_attr_c",
    srcs = ["libsrc/attr.m4"],
    outs = ["attr.c"],
    cmd = "m4 $(SRCS) > $@",
)
```
This requires that `swift-build-bazel` has `m4`, but we could also use [rules_m4](https://github.com/jmillikin/rules_m4) or maybe generate those files once and embed them in the BUILD file, similarly to `gen_config_h`.

# Testing
I've locally ran the orion-engine's tests using those BUILD files. It was only tested on linux. I've yet to test in on the CI, I'll update the PR when I'm done.

# Jira
https://swift-nav.atlassian.net/jira/software/projects/BUILD/boards/227?selectedIssue=BUILD-530


third_party/netcdf-c.BUILD
```diff
@@ -12,30 +12,983 @@
...
+        "libhdf5/nc4memcb.c",
+    ],
+    hdrs = hdrs,
+    defines = defines,
+    includes = ["include"],
+    deps = ["@hdf5"],
+)
+
+genrule(
+    name = "gen_config_h",
```
isaactorz, line 314:
```suggestion
    name = "config",
```

super nit. I think it's cleaner to not let the detail of this targets rule leak into the target name. Plus this is more consistent with other targets we have that make a `config.h` file.

third_party/netcdf-c.BUILD
```diff
@@ -12,30 +12,983 @@
...
+    outs = ["config.h"],
+    cmd = r"""
```
isaactorz, line 316: I'd like to move this to a variable so that we can `select` the string that corresponds to the platform we're building.

We'll probably need an AND chain to express the logic: https://bazel.build/docs/configurable-attributes#and-chaining

third_party/netcdf-c.BUILD
```diff
@@ -12,30 +12,983 @@
...
+/* Define to 1 if you have the `H5Z_SZIP' function. */
+#define HAVE_H5Z_SZIP 1
+
+/* Define to 1 if you have libsz */
+#define HAVE_SZ 1
+
+/* Define to 1 if you have the <hdf5.h> header file. */
+#define HAVE_HDF5_H 1
+
+/* Define to 1 if you have the <hdf5.h> header file. */
```
isaactorz, line 588: Seems like a typo on their end. I think they mean 'hdf5-hl'

third_party/netcdf-c.BUILD
```diff
@@ -12,30 +12,983 @@
...
+#define HAVE_FTW_H 1
+
+/* Define to 1 if you have the <libgen.h> header file. */
+#define HAVE_LIBGEN_H 1
+
+/* Define to 1 if you have the `strdup' function. */
+#define HAVE_STRDUP 1
+
+/* Define to 1 if you have the `strndup` function. */
+#define HAVE_STRNDUP
```
isaactorz, line 691: I wonder why these are missing the 1.

Probably doesn't matter as long as the symbol is defined.

isaactorz (commented):
Great work.

As for the `m4` generated files, what's the actual content that gets generated?

If it's not embedding info from the host system or things that are platform specific then I don't think there's any issue just adding them like we did `config.h`.

I think if we can avoid implicitly depending on m4, we should.

third_party/netcdf-c.BUILD
```diff
@@ -12,30 +12,983 @@
...
+
+defines = [
+    "HAVE_CONFIG_H",
+]
+
+cc_library(
+    name = "netcdf-c",
+    srcs = [
+        "liblib/nc_initialize.c",
+        ":dispatch",
```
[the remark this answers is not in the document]
wwieclaw: So in the netcdf all the libraries are defined as [object libraries](https://cmake.org/cmake/help/latest/command/add_library.html#id3)
```cmake
add_library(netcdf3 OBJECT ${libsrc_SOURCES})
```

And then they are composed like this:
```cmake
FOREACH(LIBS ${liblib_LIBS})
  SET(LARGS ${LARGS} $<TARGET_OBJECTS:${LIBS}>)
ENDFOREACH()

ADD_LIBRARY(netcdf nc_initialize.c ${LARGS} )
```

I tried to emulate that, but tbh I have no Idea that I did the right thing. I bet this is the reason for GCC build failure

wwieclaw:
SInce the m4 headers are huge, I did a small refactoring. I moved all the embedded source files to `third_party/netcdf-c/` directory . Each file has a corresponding `.bzl` file that contains a string constant with the entire source of the file.

third_party/netcdf-c.BUILD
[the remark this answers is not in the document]
isaactorz: Ah ok I see.

isaactorz (approved):
LGTM assuming the gcc build passes!

PR #49, opened 2023-03-09
feat: add rule to stamp binaries

isaactorz:


cc/defs.bzl
```diff
@@ -58,6 +59,22 @@ def _test_compatible_with():
         "//conditions:default": [],
     })
 
+def cc_stamped_library(name, out, template, hdrs, includes):
```
isaactorz, line 62: I think this is a more expressive name than `stamping`.

cc/defs.bzl
```diff
@@ -58,6 +59,23 @@ def _test_compatible_with():
...
+    source_name = name + "_"
+
+    stamp_file(name = source_name, out = out, template = template)
+
+    swift_cc_library(
+        name = name,
+        hdrs = hdrs,
+        includes = includes,
+        linkstamp = source_name,
+        #FIXME: Don't default to public visibility. Require users to set it.
```
isaactorz, line 75: Lets fix this before merging.

tools/stamp_file.bzl
```diff
@@ -1,7 +1,5 @@
-load("//cc:defs.bzl", "swift_cc_library")
-
-def _stamping_impl(ctx):
-    status_file = ctx.version_file
+def _stamp_file_impl(ctx):
+    status_file = ctx.info_file
```
isaactorz, line 2: there's two files that result from `workspace_status`:
- `version_file` - contains volatile status variables
- `info_file` - contains stable status variables

We should probably handle both files.

tools/stamp_file.bzl
```diff
@@ -19,8 +17,8 @@ def _stamping_impl(ctx):
 
     return [DefaultInfo(files = depset([out]))]
 
-_stamping = rule(
-    implementation = _stamping_impl,
+stamp_file = rule(
```
isaactorz, line 20: Naming

tools/stamp_file.bzl
```diff
@@ -31,20 +29,7 @@ _stamping = rule(
             executable = True,
             cfg = "exec",
             allow_files = True,
-            default = "//stamping:stamping.py",
+            default = "//tools:stamp_file.py",
```
isaactorz, line 32: Should we wrap these in a `py_binary` ?

I don't think it'll matter for caching, but seems like a good idea to use the same python toolchain as the rest of the build (right now it just finds whatevers in `/usr/bin`

PR #52, opened 2023-03-15
cc: implement c only rules [BUILD-560]

isaactorz:


isaactorz:
cc @woodfell @RReichert thought it'd be good to tag you two in this review. This essentially wraps up the API for the `swift_c` rules wrappers.

RReichert:
@isaactorz the one area that I see that might have issues with this is `gnss_converters` where were have C/C++ code in a single library, maybe some of the tests as well might face some issues.

isaactorz:
> @isaactorz the one area that I see that might have issues with this is `gnss_converters` where were have C/C++ code in a single library, maybe some of the tests as well might face some issues.

Check out this PR where I've tested: https://github.com/swift-nav/gnss-converters-private/pull/1468

But basically getting gnss-converters built correctly is what this PR addresses, since it's pretty much the only place in the codebase where we've been mixing languages.

krisukox (approved):
I tested it locally with `gnss_converters`, and it worked well. LGTM.

PR #66, opened 2023-04-05
Use pkg_zip to compress doc [BUILD-450]

krisukox:
This PR adds two more targets to the `swift_doxygen` macro:
- `target_name_html`, which filters doc files
- `target_name_zip`, which creates zip


isaactorz (approved):
LGTM!

But one question: What other files were present that we needed to filter for only html?

krisukox:
@isaactorz `Doxyfile` (doxygen template) and latex files.

PR #67, opened 2023-04-05
cc: toolchain: add support for aarch64-darwin-llvm [BUILD-563]

isaactorz:
Adds support for building with the custom llvm toolchain on aarch64-darwin.

# Design notes
- Adds `aarch64-darwin-llvm` toolchain. The `llvm-project` does not have prebuilt binaries available for clang-14 on arm macs, so we had to build this distribution from source ourselves: https://github.com/swift-nav/swift-toolchains/pull/14
- Enables treating warnings as errors on macos
- Disables `-ffp-contract=on` which was made the default in clang-14
- Enables using the llvm distribution clang-format instead of a system install for aarch64-darwin.
- note: on macos `coreutils` is an implicit dependency since we use `realpath` in the clang-format scripts.

# Testing
- Tested several of our bigger projects on m1 mac
- Tested we didn't break any compatibility on intel mac
- https://github.com/swift-nav/starling/pull/7839
- https://github.com/swift-nav/orion-engine/pull/5906
- https://github.com/swift-nav/orion/pull/5513


cc/repositories.bzl
```diff
@@ -11,18 +11,19 @@
...
     maybe(
         http_archive,
-        name = "x86_64-linux-llvm",
+        name = "aarch64-darwin-llvm",
         build_file = Label("//cc/toolchains/llvm:llvm.BUILD.bzl"),
-        url = X86_64_LINUX_LLVM,
-        strip_prefix = "clang+llvm-14.0.0-x86_64-linux-gnu-ubuntu-18.04",
-        sha256 = "61582215dafafb7b576ea30cc136be92c877ba1f1c31ddbbd372d6d65622fef5",
+        url = AARCH64_DARWIN_LLVM,
+        strip_prefix = "clang+llvm-14.0.0-arm64-apple-darwin",
```
silverjam, line 26: Sha is omitted because the build isn't final yet?

cc/toolchains/llvm/cc_toolchain_config.bzl
```diff
@@ -35,6 +35,8 @@ def cc_toolchain_config(
         "-U_FORTIFY_SOURCE",  # https://github.com/google/sanitizers/issues/247
         "-fstack-protector",
         "-fno-omit-frame-pointer",
+        # Math
```
silverjam, line 38: Should we link to something to explain why this is turned off?

cc/repositories.bzl
isaactorz: Good catch. Initially yes but I'm ready to finalize that now.

krisukox (approved):
LGTM

PR #70, opened 2023-04-12
clang-tidy: Use correct compilation flags when compiling C

krisukox:
This PR fixes the problem of analyzing C source files using `clang-tidy`. The problem was that we passed C++ flags even when compiling C source files. Clang-tidy complained about `-stdlib=libc++`.
In order to fix this, I introduced a statement that checks if the extension of the file is "c" and gets the correct compilation flags.

This issue appeared because in PR https://github.com/swift-nav/rules_swiftnav/pull/67 we enabled `-Werror` by default.

# Testing
https://github.com/swift-nav/starling/pull/7851
https://jenkins.ci.swift-nav.com/blue/organizations/jenkins/swift-nav%2Fstarling/detail/PR-7851/1/pipeline


isaactorz (approved):
Great fix.

So one thing is we were passing `-Werror` before the change (on linux at least), but maybe it wasn't getting added to the clang-tidy compile line because we were setting it as a toolchain feature?

PR #76, opened 2023-04-27
bug: fix clang format target

isaactorz:
Fixes bug from our primitive days of expecting `clang-format` to be available on the system by pointing to the version of the binary bundled with our `llvm` distribution.


peddie (approved):
I can confirm this fixed things for me locally when I pulled the change into swift-nav/orion-engine

clang_format/run_clang_format.sh
```diff
@@ -2,7 +2,7 @@
 # Usages:
 # run_clang_format format_all <CLANG_FORMAT_BIN> <CLANG_FORMAT_CONFIG>
 # run_clang_format check_file <CLANG_FORMAT_BIN> <CLANG_FORMAT_CONFIG> <INPUT> <OUTPUT>
-set -ue
+set -uex
```
isaactorz, line 5: Will remove this before merging

PR #79, opened 2023-05-08
Add gcc arm gnu 8.3 toolchain [BUILD-672]

krisukox:


platforms/BUILD.bazel
```diff
@@ -92,3 +92,13 @@ platform(
         "//cc/constraints:musl",
     ],
 )
+
+platform(
+    name = "gcc_arm_gnu_8_3",
```
isaactorz, line 97: nit: I think we should just define the platform in `starling`

isaactorz (commented):
One small nit but otherwise everything LGTM

isaactorz (approved):
lgtm!

PR #81, opened 2023-05-16
Add mkl srcs and linkopts

john-michaelburke:
## Implements
* Adds mkl static libraries to the library srcs and linkopts to fix a bug when compiling with g++.

## Testing
* Tested building run_replay in orion-engine with and without "--config=mkl"; however, the "-flto" flag for turbo is still broken for --config=system so we may need to decouple that from "--config=mkl" where it is used until we squash our own warnings and figure out a way to deal with external warnings, such as fast_csv.

### DIFF
--- a/third_party/mkl.BUILD
+++ b/third_party/mkl.BUILD
@@ -30,8 +30,18 @@ cc_import(
 
 cc_library(
     name = "mkl",
+    srcs = [
+        "@mkl//:libmkl_core.a",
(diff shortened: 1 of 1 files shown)


isaactorz (approved):
Was this change also ok with clang?

If so - LGTM

john-michaelburke:
> Was this change also ok with clang?
> 
> If so - LGTM

It appears to be working with clang as well.

PR #84, opened 2023-05-22
Add support for cross-compiling from x86_64 to aarch64 linux [CLARM-14]

krisukox:
Usage:
WORKSPACE.bazel
```
load("@rules_swiftnav//cc:repositories.bzl", "aarch64_sysroot", "register_swift_cc_toolchains", "swift_cc_toolchain")
aarch64_sysroot()
```

`bazel build --platforms=@rules_swiftnav//platforms:aarch64_bullseye //...`

This PR also adds support for cross-compilation for openssl.

### DIFF
--- a/cc/toolchains/llvm/cc_toolchain_config.bzl
+++ b/cc/toolchains/llvm/cc_toolchain_config.bzl
@@ -28,6 +28,8 @@ def cc_toolchain_config(
         target_system_name,
         builtin_sysroot = None,
         is_darwin = False):
+    cross_compile = host_system_name != target_system_name
+
     # Default compiler flags:
     compile_flags = [
         "--target=" + target_system_name,
@@ -72,9 +74,8 @@ def cc_toolchain_config(
     cxx_flags = [
         # The whole codebase should build with c++14
         "-std=c++14",
-        # Use bundled libc++ for hermeticity
-        "-stdlib=libc++",
-    ]
+        # Use bundled libc++ for hermeticity if not cross compiling
+    ] + ["-stdlib=libstdc++"] if cross_compile else ["-stdlib=libc++"]
 
     link_flags = [
         "--target=" + target_system_name,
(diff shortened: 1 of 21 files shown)


cc/toolchains/llvm/cc_toolchain_config.bzl
```diff
@@ -28,6 +28,8 @@ def cc_toolchain_config(
         target_system_name,
         builtin_sysroot = None,
         is_darwin = False):
+    cross_compile = host_system_name != target_system_name
```
isaactorz, line 31: This is a little dangerous.

We don't enforce any sort of convention around what `host_system_name` and `target_system_name` are.

This led to `-stdlib=libstdc++` being selected because they didn't match in the `x86_64-linux` config, even though we weren't cross compiling.

What we should probably do is define a set of constants for the clang target triplets (`x86_64-linux-unknown-linux-gnu`, `aarch64-unknown-linux-gnu`, etc..)

Then in this function we should write a quick check to make sure we're using one of those constants, otherwise it's an error.

That would make this a lot more robust.

platforms/BUILD.bazel
```diff
@@ -101,3 +101,11 @@ platform(
         "//cc/constraints:yocto_generic_toolchain",
     ],
 )
+
+platform(
+    name = "aarch64_linux",
```
isaactorz, line 106: Let's name this platform `bullseye_aarch64` (or maybe `aarch64_bullseye`, what do you think?).

We should add an extra constraint value to distinguish it.

I'm thinking either introduce a new `constraint_setting` for `glibc` version, or for `debian`, and then the `constraint_values` could either be `glibc_2_2x` or `bullseye` respectively.

krisukox: Added glibc `constraint_value` for libc `constraint_setting`.

isaactorz (approved):
I did a build on x86_64-darwin and aarch64-darwin just as a sanity check.

LGTM!

PR #85, opened 2023-05-24
Add aarch64_linux support for netcdf-c

wwieclaw:
I generated a config for aarch64_linux is the same as the one for x86_64_linux, so I just reused it.

# Testing
orion built just fine on my x84_64 machine with this commit


isaactorz (approved):
LGTM

PR #89, opened 2023-06-06
add tags=["manual"] to image_tag rules

wwieclaw:
without this change `image_stamp_tag` is built by default, even when stamping is not enabled and it cannot find the `stable-status.txt` file


isaactorz (approved):

isaactorz:
@wwieclaw I was testing this and I've also added the `manual` tag to the image transition. Otherwise you get a surprise x86_64/aarch64 build when using `//...`

isaactorz:
@wwieclaw we should probably have also considered just passing through `**kwargs` so that users can decide whether to add the `manual` tag...

PR #91, opened 2023-06-07
add kwargs to image_tar rules

wwieclaw:


silverjam:
What is this for?  Can we fill out the PR description and add a Jira link (if it exists)?

silverjam:
Cc @isaactorz @wwieclaw

isaactorz:
> What is this for? Can we fill out the PR description and add a Jira link (if it exists)?

We had some discussion in a previous PR when initially creating this macro about whether it would be better to set the `manual` tag by default, or let that be something users of the API specify themselves.

This PR was a follow up to that discussion so there's no jira.

isaactorz:
Gonna close for now so this doesn't keep pinging the review queue.

I think we can leave it as is for the time being.

PR #92, opened 2023-06-13
cc: toolchain: Add x86_64 sysroot [CLARM-39]

krisukox:
This PR adds ability to use x86_64 sysroot with llvm toolchain. Usage:
`bazel build //...  --@rules_swiftnav//cc:enable_sysroot=true`

**Related PR**
https://github.com/swift-nav/swift-toolchains/pull/19

### DIFF
--- a/cc/toolchains/llvm/cc_toolchain_config.bzl
+++ b/cc/toolchains/llvm/cc_toolchain_config.bzl
@@ -35,7 +35,7 @@ def cc_toolchain_config(
     if not is_target_triplet(target_system_name):
         fail(target_system_name + " is not a target tripplet")
 
-    cross_compile = host_system_name != target_system_name
+    use_bundled_libcpp = builtin_sysroot != None
 
     # Default compiler flags:
     compile_flags = [
@@ -82,7 +82,7 @@ def cc_toolchain_config(
         # The whole codebase should build with c++14
         "-std=c++14",
         # Use bundled libc++ for hermeticity if not cross compiling
-    ] + ["-stdlib=libstdc++"] if cross_compile else ["-stdlib=libc++"]
+    ] + ["-stdlib=libstdc++"] if use_bundled_libcpp else ["-stdlib=libc++"]
 
     link_flags = [
         "--target=" + target_system_name,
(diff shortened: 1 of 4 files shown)


cc/toolchains/llvm/cc_toolchain_config.bzl
```diff
@@ -35,7 +35,7 @@ def cc_toolchain_config(
     if not is_target_triplet(target_system_name):
         fail(target_system_name + " is not a target tripplet")
 
-    cross_compile = host_system_name != target_system_name
+    use_bundled_libcpp = builtin_sysroot != None
```
isaactorz, line 38: This won't quite work because the mac toolchains use the `builtin_sysroot` argument: https://github.com/swift-nav/rules_swiftnav/blob/af0ddfab4a282d642f00d4665aa31e342b4eaef3/cc/toolchains/llvm/aarch64-darwin/BUILD.bazel#L96 (and we still want to use libc++ in that context).

Maybe:
```suggestion
    use_bundled_libcpp = builtin_sysroot != None && !is_darwin
```

cc/toolchains/llvm/x86_64-linux/BUILD.bazel
```diff
@@ -62,14 +74,18 @@ filegroup(
...
+        "//conditions:default": [],
+    }),
 )
 
 filegroup(
     name = "objcopy_files",
     srcs = [
         ":wrappers",
         "@x86_64-linux-llvm//:objcopy",
+        "@x86_64-sysroot",
```
isaactorz, line 88: Should we have a select here?

isaactorz (commented):
Would also like a link to a test PR I can run so we can test these changes on mac.

cc/toolchains/llvm/cc_toolchain_config.bzl
```diff
@@ -35,7 +35,7 @@ def cc_toolchain_config(
     if not is_target_triplet(target_system_name):
         fail(target_system_name + " is not a target tripplet")
 
-    cross_compile = host_system_name != target_system_name
+    use_bundled_libcpp = builtin_sysroot != None and not is_darwin
```
isaactorz, line 38:
```suggestion
    use_libstdcpp = builtin_sysroot != None and not is_darwin
```

nit - the variable name doesn't match the logic. When this is true we should use `libstdc++` not `libc++`

isaactorz (approved):
LGTM!

PR #93, opened 2023-06-13
add zstd build file

wwieclaw:
tested in https://github.com/swift-nav/albatross/pull/433


third_party/zstd.BUILD
```diff
@@ -0,0 +1,85 @@
...
+        "lib/decompress/zstd_ddict.h",
+        "lib/decompress/zstd_decompress_block.h",
+        "lib/decompress/zstd_decompress_internal.h",
+        "lib/dictBuilder/cover.h",
+        "lib/dictBuilder/divsufsort.h",
+        "lib/zdict.h",
+        "lib/zstd.h",
+        "lib/zstd_errors.h",
+    ],
+    copts = ["-fpic"],
```
isaactorz, line 76: Can we remove this?

To me this is something we should enable as part of a build configuration instead of a specific library (we probably aren't linking this shared anyway).

PR #95, opened 2023-06-16
make it more obvious that clang-tidy is reading a custom config

silverjam:


silverjam:
@isaactorz Any reason not to implement loading the config file like this?

isaactorz:
> 
> @isaactorz Any reason not to implement loading the config file like this?

I don't think so. Definitely makes it more obvious.

I think it's a good change but I'd let @krisukox have the final say since we set most of this up.

krisukox:
Yes, it makes much more sense. I did it that way because this flag was introduced in `clang-tidy` 14, and we wanted to support other versions too.

The only thing is that it will change a bit how `clang-tidy` works. Now it looks for the `.clang-tidy` config in the root directory of a project, and if it can't find it, `clang-tidy` uses the config from the `rules_swiftnav/clang-tidy/.clang-tidy` file. After the change, `clang-tidy` will always use the config from the `rules_swiftnav/clang-tidy/.clang-tidy` file. We can, of course, specify a different config using the `--@rules_swiftnav//clang_tidy/clang_tidy_config=<config>` flag.

If we want to keep the current functionality, I would go with something like this:
```
if [ ! -f ".clang-tidy" ]; then
    "${CLANG_TIDY_BIN}" --config-file=$CLANG_TIDY_CONFIG "$@"
else
    "${CLANG_TIDY_BIN}" "$@"
fi
```

PR #100, opened 2023-07-12
Do not run clang-format if no files are selected

krisukox:
Use the `-r` switch with xargs to avoid running clang-format if no files are provided.


isaactorz (approved):
Nice

PR #101, opened 2023-07-12
cc: enable symbolized output in sanitizer builds [BUILD-723]

isaactorz:
Enables properly symbolized sanitizer output with clang.

Clang requires the `llvm-symbolizer` tool to convert addresses to source locations (see: https://clang.llvm.org/docs/AddressSanitizer.html#symbolizing-the-reports).

It either needs to be on the path or set via the `ASAN_SYMBOLIZER_PATH` environment variable.

# Design Notes

We can use the `cc*` `env` attr and [make variable substitution](https://bazel.build/reference/be/make-variables#predefined_label_variables) to make this work hermetically under bazel.

Our custom built llvm distribution for apple silicon is currently missing the extra clang tools so this will only work on intel mac and linux for now.

# Testing

Test PR:
- https://github.com/swift-nav/orion-engine/pull/6680

### DIFF
--- a/cc/defs.bzl
+++ b/cc/defs.bzl
@@ -591,7 +591,8 @@ def swift_cc_test(name, type, **kwargs):
     kwargs["tags"] = [TEST, type] + kwargs.get("tags", [])
     kwargs["target_compatible_with"] = kwargs.get("target_compatible_with", []) + _test_compatible_with()
     kwargs["env"] = select({
-        Label("//cc:_enable_symbolizer"): {"ASAN_SYMBOLIZER_PATH": "$(location @x86_64-linux-llvm//:symbolizer)"},
+        Label("//cc:_enable_symbolizer_x86_64_linux"): {"ASAN_SYMBOLIZER_PATH": "$(location @x86_64-linux-llvm//:symbolizer)"},
+        Label("//cc:_enable_symbolizer_x86_64_darwin"): {"ASAN_SYMBOLIZER_PATH": "$(location @x86_64-darwin-llvm//:symbolizer)"},
         "//conditions:default": {},
     })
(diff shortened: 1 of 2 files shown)


ljbade (approved):
im not 100% on top of bazel yet but this looks like a simple change

cc/defs.bzl
```diff
@@ -99,6 +99,22 @@ def _create_hdrs(**kwargs):
...
+        # clever here.
+        Label("//cc:_enable_symbolizer_x86_64_linux"): dict(val, **{"ASAN_SYMBOLIZER_PATH": "$(location @x86_64-linux-llvm//:symbolizer)"}),
+        Label("//cc:_enable_symbolizer_x86_64_darwin"): dict(val, **{"ASAN_SYMBOLIZER_PATH": "$(location @x86_64-darwin-llvm//:symbolizer)"}),
+        "//conditions:default": {},
+    })
+
+def _symbolizer_data():
+    return select({
+        Label("//cc:_enable_symbolizer_x86_64_linux"): ["@x86_64-linux-llvm//:symbolizer"],
+        Label("//cc:_enable_symbolizer_x86_64_darwin"): ["@x86_64-darwin-llvm//:symbolizer"],
```
silverjam, line 114: Do we need a setting for aarch64?

isaactorz: We don't.

We built the llvm distribution for apple silicon from source since there were no pre-built binaries available. I couldn't figure out how to build the extra clang tools (clang-tidy, llvm-symbolizer, etc..)

So this is currently only supported on intel mac and linux.

silverjam: :+1:

silverjam (approved):

PR #104, opened 2023-08-21
Make static linking default option for Musl toolchain

jithk:
Part of the efforts to add a perf build for profiling

https://swift-nav.atlassian.net/browse/SCD-245

~This PR will add a musl-static toolchain for ARM. Just copies over the standard musl toolchain and add options for static build~

After discussing with @isaactorz , we will just make static linking default for all musl toolchain as this is probably the expected behaviour.


silverjam (commented):
LGTM, deferring to @isaactorz for approval

isaactorz (commented):
For musl toolchains static linking is implicit - you should automatically get a fully statically linked binary.

So this toolchain is probably redundant.

Also probably worth mentioning that when you write a toolchain from scratch you have to manually specify what flags are activated for `opt` `dbg` - like we do here for instance: https://github.com/swift-nav/rules_swiftnav/blob/a38ff0166ba73f80cabf41acdcd70d0bc50dc7f0/cc/toolchains/gcc_arm_gnu_8_3/config.bzl#L94 (I'm assuming you want to use this with some sort of optimization enabled)

isaactorz:
> 
> Actually - let me double check if that's the case for the c++ standard library as well as the c standard library

Sorry @jithk I was wrong about that.

I've mostly used musl toolchains in the context of rust compilation. They must tell the linker to create a fully statically linked binary when targeting musl libc.

It does indeed look like C/C++ binaries are still dynamically linked to the interpreter without adding the `-static` flag.

I think that we should just add that to the default link options of the existing musl toolchains, since that's probably the more expected behavior.

jithk:
> I think that we should just add that to the default link options of the existing musl toolchains, since that's probably the more expected behavior.

Updated

isaactorz (commented):
Looks good but I'm guessing you'll probably also want to add the ability to enable optimizations with `-c opt` correct?

This is how to enable that for a custom toolchain: https://github.com/swift-nav/rules_swiftnav/blob/e5f9de410d5081c55bd0c18127798f9f1d1cb37c/cc/toolchains/gcc_arm_gnu_8_3/config.bzl#L94

jithk:
> Looks good but I'm guessing you'll probably also want to add the ability to enable optimizations with `-c opt` correct?
> 
> This is how to enable that for a custom toolchain:

added :+1:

isaactorz (dismissed):
lgtm!

isaactorz (approved):
LGTM - I added the changes you've made to the other two musl toolchains.

PR #107, opened 2023-08-29
feat: clang-format: add format_all variant [DIP-243]

isaactorz:
For our clang-format utility - we use git to obtain a list of files to run the formatter against.

To speed up formatting and linting we attempt to get a subset of the projects files. We currently use `git describe --tags`. This is a reasonable heuristic - but it will not work properly if the project does not have tags.

This PR adds a rule that will use `git ls-files` instead for projects that do not have tags at the sacrifice of speed.

# Testing

- Tested in albatross where the issue was discovered: https://github.com/swift-nav/albatross/pull/451

### DIFF
--- a/clang_format/run_clang_format.sh
+++ b/clang_format/run_clang_format.sh
@@ -12,6 +12,17 @@ format_all() {
         echo ".clang-format file not found. Bazel will copy the default .clang-format file."
         cp $CLANG_FORMAT_CONFIG .
     fi
+    git ls-files '*.[ch]' '*.cpp' '*.cc' '*.hpp' | xargs -r $CLANG_FORMAT_BIN -i
+}
+
+format_diff() {
+    CLANG_FORMAT_CONFIG=$(realpath $1)
+
+    cd $BUILD_WORKSPACE_DIRECTORY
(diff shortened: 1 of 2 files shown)


clang_format/run_clang_format.sh
```diff
@@ -7,6 +7,17 @@ set -ue
 format_all() {
     CLANG_FORMAT_CONFIG=$(realpath $1)
 
+    cd $BUILD_WORKSPACE_DIRECTORY
+    if ! test -f .clang-format; then
+        echo ".clang-format file not found. Bazel will copy the default .clang-format file."
+        cp $CLANG_FORMAT_CONFIG .
+    fi
+    git ls-files '*.[ch]' '*.cpp' '*.cc' '*.hpp' | xargs -r $CLANG_FORMAT_BIN -i
```
silverjam, line 15: Should we have `*.cxx` `*.hxx` for completeness?

silverjam (approved):

clang_format/run_clang_format.sh
isaactorz: Wouldn't hurt!

PR #108, opened 2023-09-15
cc: toolchain: enable external_include_paths feature on mac

isaactorz:
# Description

We require the external include paths feature to be able to compile our code locally with `-I` to ensure compiler and linter warnings are picked up in header files, and with `-isystem` when the project is used externally to silence warnings when building with the consuming projects options.

For some reason this feature is only enabled for the linux cc toolchain configuration in the part of the toolchain config we reuse from bazel (unix_cc_toolchain_config): https://github.com/bazelbuild/bazel/blob/820ed9bf86bd8c3ced8aa65784be3f01e40414ec/tools/cpp/unix_cc_toolchain_config.bzl#L1373

# Design Notes

The only solution here is to vendor the `unix_cc_toolchain_config` implementation and make the custom edits we require. Specificall we need to add the `external_include_paths_feature` to the macos toolchain configuration - as I've done here.

Eventually my goal has always been to decouple from `unix_cc_toolchain_config` so we can add new toolchain features, this can just be considered accelerating that plan and opens up several possibilities for improving our build system in the future.

The main risk this introduces is drift from the autoconfigured cc toolchain. The implementation is coupled to some degree to internal bazel API's so we will have to take care when upgrading bazel versions or making changes to make sure our implementation stays compatible. Here I've copied the implementation as of `6.2.0`, which is our current bazel version. We still need the autoconfigured toolchain for supporting building with GCC, and arbitrary compilers in general.

In the case of mac - this change implies that our code will no longer build with the autoconfigured toolchain on that platform. I see no reason to continue to support buildling with Apple Clang since we have set up the hermetic toolchains. We can however open a PR to see if this change can be merged upstream.

# Testing
- Test PR: https://github.com/swift-nav/orion-engine/pull/7116

I've also confirmed this is working as expected on intel and m1 macs.

### DIFF
--- a/cc/toolchains/llvm/cc_toolchain_config.bzl
+++ b/cc/toolchains/llvm/cc_toolchain_config.bzl
@@ -9,7 +9,7 @@
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
 load(
-    "@bazel_tools//tools/cpp:unix_cc_toolchain_config.bzl",
+    "unix_cc_toolchain_config.bzl",
     unix_cc_toolchain_config = "cc_toolchain_config",
 )
 load("//cc/toolchains/llvm:target_triplets.bzl", "is_target_triplet")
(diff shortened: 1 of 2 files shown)


silverjam:
> We can however open a PR to see if this change can be merged upstream.

Would it be a lot of effort to submit an upstream PR?

silverjam (approved):

isaactorz:
> > We can however open a PR to see if this change can be merged upstream.
> 
> Would it be a lot of effort to submit an upstream PR?

Not at all, I'll open one next week.

Not sure if they'd patch an existing release though (not sure if they'll consider it a bug).

But that would enable us to get back to compatible in the future when this releases upstream.

PR #109, opened 2023-09-19
Add build file for suitesparse's CHOLMOD / SPQR [OC-310]

peddie:


third_party/suitesparse.BUILD
```diff
@@ -0,0 +1,600 @@
...
+    srcs = [
+        "SuiteSparse_config/SuiteSparse_config.c",
+    ],
+    hdrs = [
+        "SuiteSparse_config/SuiteSparse_config.h",
+    ],
+    includes = [
+        "SuiteSparse_config/",
+    ],
+    strip_include_prefix = "SuiteSparse_config",
```
isaactorz, line 18: For third party code we should try to avoid using `strip_include_prefix` which causes the include path to be propagated with `-I` - The `includes` statements above should be sufficient.

third_party/suitesparse.BUILD
```diff
@@ -0,0 +1,600 @@
...
+    linkopts = [
+        "-lm",
+        "-lpthread",
+        "-lblas",
```
isaactorz, line 22: forewarning - for targeting arm we build against a sysroot so these packages might need to be added there

third_party/suitesparse.BUILD
```diff
@@ -0,0 +1,600 @@
...
+        "-llapack",
+    ],
+    visibility = ["//visibility:public"],
```
isaactorz, line 25: Since you declared a default of public visibility at the top - all of these are redundant.

third_party/suitesparse.BUILD
```diff
@@ -0,0 +1,600 @@
...
+        "-Iexternal/suitesparse/AMD/Source",
+    ],
+    hdrs = [
+        "AMD/Include/amd.h",
+        "AMD/Include/amd_internal.h",
+    ],
+    includes = [
+        "AMD/Include",
+    ],
+    strip_include_prefix = "AMD/Include",
```
isaactorz, line 68: Same feedback as above and for the rest of the targets.

third_party/suitesparse.BUILD
```diff
@@ -0,0 +1,600 @@
...
+        "-lblas",
+        "-llapack",
+    ],
+    visibility = ["//visibility:public"],
+)
+
+cc_library(
+    name = "GKlib",
+    defines = [
+        "LINUX",
```
isaactorz, line 187: Does this imply this dependency only works on linux?

third_party/suitesparse.BUILD
```diff
@@ -0,0 +1,600 @@
...
+        "CHOLMOD/SuiteSparse_metis/GKlib/util.c",
+        # These fail to compile as of 2023 -- Tim Davis commented out
+        # various definitions that apparently are not needed.
+        #
+        # "CHOLMOD/SuiteSparse_metis/GKlib/graph.c",
+        # "CHOLMOD/SuiteSparse_metis/GKlib/string.c",
+        # "CHOLMOD/SuiteSparse_metis/GKlib/timers.c",
+
+    ],
+    textual_hdrs = [
```
isaactorz, line 241: Empty attribute can be removed.

isaactorz (commented):
You should be able to run `bazel run //:buildifier` to format the code.

third_party/suitesparse.BUILD
peddie: Good question.  This value comes from https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/dev/CHOLMOD/SuiteSparse_metis/GKlib/GKlibSystem.cmake#L24 which defines `LINUX` as long as neither `MSVC` nor `MINGW` matches.  In spite of this, as far as I know, there's no problem running CHOLMOD/METIS on Macintosh.

third_party/suitesparse.BUILD
peddie: I'm afraid I don't totally understand what that implies, either in general or for this PR.

third_party/suitesparse.BUILD
peddie: Great -- while developing, I thought I needed this for e.g. `cholmod` to correctly get headers from `amd` etc., but it looks like you're right.

silverjam:
cc @isaactorz

isaactorz:
> cc @isaactorz

There was some discussion here that moved to slack but the gist of it is we'll need to update our sysroot to include the new dependencies on `blas` and `lapack`.

Matt's switched to a different task temporarily though.

silverjam:
> > cc @isaactorz
> 
> There was some discussion here that moved to slack but the gist of it is we'll need to update our sysroot to include the new dependencies on `blas` and `lapack`.
> 
> Matt's switched to a different task temporarily though.

Thanks for the update.

silverjam:
I'm going to remove the review request for now

third_party/suitesparse.BUILD
```diff
@@ -300,6 +312,8 @@ cc_library(
         "CHOLMOD/SuiteSparse_metis/include/",
     ],
     linkopts = [
+        "-Lexternal/aarch64-sysroot/usr/lib/aarch64-linux-gnu/lapack",
```
isaactorz, line 315: I thought about this a little more and we're probably going to need to guard these in a `select` statement so that they're only put on the link line when we're actually using the sysroot.

The way bazel's build tree is structured - the `external/aarch64-sysroot/` folder may be on the path even when not doing a build against a sysroot. This could cause the build to attempt to link libs for the wrong platform.

isaactorz: I've added the appropriate guards.

isaactorz (approved):
LGTM assuming all testing goes well!

third_party/suitesparse.BUILD
peddie: Thanks.

peddie:
I've updated to your version of this PR (with `b6e2417` at the top) but now get errors when testing the `x86_64` sysroot in `orion-engine`:

```
ERROR: /home/peddie/orion-engine/BUILD.bazel:1159:10: Linking latlon2areaid failed: (Exit 1): clang failed: error executing command (from target //:latlon2areaid) external/rules_swiftnav/cc/toolchains/llvm/x86_64-linux/wrappers/clang @bazel-out/k8-fastbuild/bin/latlon2areaid-2.params

Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
ld.lld: error: unable to find library -lblas
ld.lld: error: unable to find library -llapack
```

PR #110, opened 2023-09-29
feat: add local archive utility

isaactorz:
Adds a handy `local_archive` utility.

See: https://www.stevenengelhardt.com/2023/03/01/practical-bazel-local_archive-workspace-rule/

Related: https://github.com/swift-nav/swift-toolchains/pull/22


peddie (approved):
Seems OK to me

PR #111, opened 2023-09-29
Optionally configure the input dir [OI-652]

IsakTjernberg:
### Jira ticket
https://swift-nav.atlassian.net/browse/OI-652

### Background
The companion PR in orion-engine https://github.com/swift-nav/orion-engine/pull/7170 needs these changes in order to configure INPUT_DIR and PROJECT_NAME in the `Doxygen.in` file in orion-engine repo.
This Bazel rule is also used from Starling, but the `Doxygen.in` file in Starling does not have the `@INPUT_DIR@` or `@PROJECT_NAME@` tags so the replace with `sed` should not have any effect there.

### Testing
- [x] Tested in orion-engine: https://github.com/swift-nav/orion-engine/pull/7170
- [x] Smoke tested in Starling: https://github.com/swift-nav/starling/pull/8658

### DIFF
--- a/doxygen/doxygen.bzl
+++ b/doxygen/doxygen.bzl
@@ -20,6 +20,10 @@ def _swift_doxygen_impl(ctx):
 
     config = configure_file_impl(ctx, vars, ctx.attr.name + "_Doxyfile")[0].files.to_list()[0]
 
+    input_dir = ""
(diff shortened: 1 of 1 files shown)


isaactorz (approved):
LGTM.

PR #112, opened 2023-10-04
fix: various mac build fixes and improvements [DIP-311]

isaactorz:
The recent macos sdk v15 update has caused a number of issues with our builds that require various tweaks to our build settings to fix.

## Switch to ninja for cmake generator

The `cmake` build process involves bootstrapping a gnu make binary from source. The update has caused some strange build errors during this process. More info can be found in the ticket I've created here: https://github.com/bazelbuild/rules_foreign_cc/issues/1099

In the meantime we can just switch to using the ninja generator.

## Drop support for shared linking

We make liberal use of "weak linking" for various purposes across the codebase - usually implemented using function pointers for (I assume) portability.

However, this means that most of our libraries can't really be built as shared libraries due to resulting "undefined reference" errors.

In bazel the build system will by default always try to build the shared library unless `linkstatic` is set to `True`. So on mac this was causing errors during the normal build. Not sure why linux was not having problems (all our CMake builds fail when `-DBUILD_SHARED_LIBS=TRUE`).

To fix this I added `-undefined dynamic_lookup` to the link flags on mac to tell the linker we'll supply the symbols later at the final link step.

Now with the update, for whatever reason these flags appear to be causing the following crash on m1 macs when the linker tries to link a binary:
```
Use --sandbox_debug to see verbose messages from the sandbox and retain the sandbox build root for debugging
0  0x102277648  __assert_rtn + 72
1  0x10224c6c8  ld::AtomFileConsolidator::addAtomFile(ld::AtomFile const*, ld::AtomFile const*, bool) + 4860
2  0x1022581d0  ld::AtomFileConsolidator::addAtomFile(ld::AtomFile const*) + 148
3  0x102275b20  ld::pass::stubs(ld::Options const&, ld::AtomFileConsolidator&) + 1644
4  0x10225f84c  ld::AtomFileConsolidator::resolve() + 12232
5  0x1021ea1a8  main + 9104
ld: Assertion failed: (slot < _sideTableBuffer.size()), function addAtom, file AtomFileConsolidator.cpp, line 2158.
clang: error: linker command failed with exit code 1 (use -v to see invocation)
INFO: Elapsed time: 190.377s, Critical Path: 36.73s
INFO: 1484 processes: 210 internal, 1274 darwin-sandbox.
FAILED: Build did NOT complete successfully
```

Since we've never really supported shared linking except in some exceptional cases, I've disabled shared library builds by default across all our code and thus can remove the problematic flags (and probably get a perf boost as well).

## Make the minimum macos target consistent

Another annoyance with the mac builds was that the bazel code was getting compiled with a different macos deployment target than the cmake libraries, which resulting in the build output getting spammed with linker warnings referencing this difference.

On my system the bazel target was `13.0`, while the cmake value was `13.6`. I'm not sure how (as of our current version of bazel 6.2) this value is computed.

In CMake it's interpolated from the system.

To fix this I've wired up our toolchain to respect the [macos_minimum_os](https://bazel.build/reference/command-line-reference#flag--macos_minimum_os) flags and add `-mmacos-version-min` to our build. This is just taken from the most recent `HEAD` of `bazel` which is not in 6.x.

The CMake side requires explicitly setting [CMAKE_OSX_DEPLOYMENT_TARGET](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html) variable to the same value.

It would be better if `rules_foreign_cc` also respected the `macos_minimum_os` flag so the value would not have to be hardcoded. We can perhaps open a ticket to request this.

`13.0` seems like a reasonable minimum deployment target since these builds are only intended for development machines.

## Testing

I've tested these changes on m1 and intel macs as well as the following test PR's:
- https://github.com/swift-nav/starling/pull/8677
- https://github.com/swift-nav/orion-engine/pull/7239

### DIFF
--- a/third_party/check.BUILD
+++ b/third_party/check.BUILD
@@ -22,6 +22,10 @@ cmake(
         "CMAKE_INSTALL_LIBDIR": "lib",
         "HAVE_SUBUNIT": "0",
     },
+    generate_args = [
+        "-GNinja",
+        "-DCMAKE_OSX_DEPLOYMENT_TARGET=13.0",
+    ],
     lib_source = ":srcs",
     linkopts = select({
         "@bazel_tools//src/conditions:darwin": ["-lpthread"],
@@ -30,10 +34,6 @@ cmake(
             "-lrt",
         ],
     }),
-    generate_args = [
-        "-GNinja",
-        "-DCMAKE_OSX_DEPLOYMENT_TARGET=13.0",
-    ],
     out_static_libs = select({
         "@bazel_tools//src/conditions:windows": ["check.lib"],
         "//conditions:default": ["libcheck.a"],
--- a/third_party/check.BUILD
+++ b/third_party/check.BUILD
@@ -30,6 +30,10 @@ cmake(
             "-lrt",
         ],
     }),
+    generate_args = [
+        "-GNinja",
+        "-DCMAKE_OSX_DEPLOYMENT_TARGET=13.0",
+    ],
     out_static_libs = select({
         "@bazel_tools//src/conditions:windows": ["check.lib"],
         "//conditions:default": ["libcheck.a"],
--- a/cc/BUILD.bazel
+++ b/cc/BUILD.bazel
@@ -55,6 +55,18 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+bool_flag(
+    name = "enable_shared",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "_enable_shared",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)
+
 # Disable warnings as errors with --@rules_swiftnav//cc:warnings_as_errors=false
 bool_flag(
     name = "warnings_as_errors",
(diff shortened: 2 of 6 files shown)


third_party/check.BUILD
```diff
@@ -22,6 +22,10 @@ cmake(
         "CMAKE_INSTALL_LIBDIR": "lib",
         "HAVE_SUBUNIT": "0",
     },
+    generate_args = [
+        "-GNinja",
```
silverjam, line 26: Wouldn't this require all macOS devs to have Ninja preinstalled on their system?

isaactorz: Let me double check.

I assumed that since they go through the trouble of bootstrapping `gnu make` that they also provide a ninja binary or something.

But to be fair I didn't notice if one was getting downloaded/built.

isaactorz: I checked each platform and the build scripts for the cmake builds all reference  ninja instances that live in the `execroot`

Looks like pre-built instance on x86_64-darwin and x86_64-linux,
and built from source on aarch64-darwin.

silverjam: 👍

silverjam (approved):
Pending anymore testing you'd like to do

isaactorz:
Test PR's are matching so I'm going to merge

PR #115, opened 2023-10-04
Add iec559 double definition check

armallen:
# Rationale

We need to check double type definition for ASIL purposes. This adds a simple check, then called by integrity_core_common (or any library needing to check double definition)

# Related PRs
https://github.com/swift-nav/cmake/pull/169
https://github.com/swift-nav/orion-engine/pull/7226


isaactorz (commented):
I'm a little confused about how this is supposed to work.

I'm actually surprised that the build didn't complain about making a `cc_binary` a dep of `cc_library`.

In any case this is a runtime failure so I don't see how this will fail the build if the compiler is indeed not compatible.

Can we make this check `constexpr` instead so that it actually fails at compile time?

armallen:
> I'm a little confused about how this is supposed to work.
> 
> I'm actually surprised that the build didn't complain about making a `cc_binary` a dep of `cc_library` - that's probably the first warning sign.
> 
> In any case this is a runtime failure so I don't see how this will fail the build if the compiler is indeed not compatible.
> 
> Can we make this check `constexpr` instead so that it actually fails at compile time?

🤦  sorry about this, I think we'll go for a static assert in the library instead

PR #117, opened 2023-10-06
Try to fix paths in clang-tidy output

woodfell:


woodfell:
@isaactorz as discussed in the build sync. Probably not an idiomatic solution but it solves my problem and makes the output of clang-tidy useable by other tooling

isaactorz:
> @isaactorz as discussed in the build sync. Probably not an idiomatic solution but it solves my problem and makes the output of clang-tidy useable by other tooling

pinging myself to remember to take a look at it.

At first glance seems perfectly reasonable just want to take a closer look before approving.

clang_tidy/run_clang_tidy.sh
```diff
@@ -22,6 +22,7 @@ fi
 touch $OUTPUT
 truncate -s 0 $OUTPUT
 
-"${CLANG_TIDY_BIN}" "$@"
+"${CLANG_TIDY_BIN}" "$@" | sed "s;.*execroot/[^/]*/;;"
```
isaactorz, line 25: nit: can we leave a small comment here as a reminder of what this was for?

isaactorz (approved):

PR #118, opened 2023-10-06
cc: quiet builds

isaactorz:
Removes noise in our builds so we can focus on warnings and issues that matter.

# Design Notes

## Remove default warnings from toolchain

The OTB bazel cc toolchain adds a few warnings to the compile line for every file (i.e. `-Wall`). 
We've kept this convention but it has made our builds extremely noisy, particularly when building 
third party code.

I've moved these options to only be added in our `swift_cc` wrappers.

## Silence implicit warnings when building third party code

Clang has a handful of warnings that are always on by default. I've added a new feature to our 
toolchain that disables these warnings when building third party code so that builds aren't 
spammed by warnings we're not likely to ever fix.

This has the downside of masking poor code hygiene in our dependencies, but
the option can be overridden at the command line:
`bazel build --features=-no_default_warnings //...`

## Silence warnings about no symbols on macos

Libtool outputs a warning message if a file has no symbols. This shows up quite
frequently in our builds - for instance in cases like `boringssl` where entire files
are `#ifdef'd` away due to platform.

This is something we can ignore.

## Silence warnings about duplicate libraries on macos

This one is new with the latest linker upgrade. Now `ld` will issue a warning if
for instance `-lpthread -lpthread` ends up on the link line.

This duplication is unavoidable in a big c++ codebase as you can't assume someone
else is going to provide the dependency for you.

## Silence -Wno-deprecated-declarations in tests

This mostly comes from the extensive use of deprecated api's in gtest.
I've added a flag so that this can be re-enabled locally.

# Testing

I've tested the changes on both darwin platforms and linux and the difference is
night and day.

The two following test PR's should satisfy that no major regressions have been introduced:
- https://github.com/swift-nav/starling/pull/8678
- https://github.com/swift-nav/orion-engine/pull/7243

### DIFF
--- a/cc/BUILD.bazel
+++ b/cc/BUILD.bazel
@@ -29,6 +29,19 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+# Enable warnings for deprecated declarations with --@rules_swiftnav//cc:warn_deprecated_declarations=true
+bool_flag(
+    name = "warn_deprecated_declarations",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
+)
+
+config_setting(
+    name = "_warn_deprecated_declarations",
+    flag_values = {":warn_deprecated_declarations": "true"},
+    visibility = ["//visibility:public"],
+)
+
 # Enable exceptions with --@rules_swiftnav//cc:enable_exceptions=true
 bool_flag(
     name = "enable_exceptions",
--- a/cc/BUILD.bazel
+++ b/cc/BUILD.bazel
@@ -29,16 +29,16 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
-# Enable warnings for deprecated declarations with --@rules_swiftnav//cc:warn_deprecated_declarations=true
+# Enable warnings in tests for deprecated declarations with --@rules_swiftnav//cc:tests_warn_deprecated_declarations=true
 bool_flag(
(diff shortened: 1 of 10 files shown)


silverjam (approved):

cc/BUILD.bazel
```diff
@@ -29,6 +29,19 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
+# Enable warnings for deprecated declarations with --@rules_swiftnav//cc:warn_deprecated_declarations=true
+bool_flag(
```
woodfell, line 33: I like the PR overall, but this is making me nervous. This is exactly the thing that needs to be really annoying so it doesn't get forgotten and we end up in a situation where an API gets removed but we haven't updated code.

How about altering the `swift_cc_test` target so that it doesn't make use of this option, but all other target types do? It seems reasonable for unit tests to be interacting with deprecated and soon-to-be removed APIs, but we should remain annoyingly noisy on production code

isaactorz: Completely agree.

I've only actually wired this config setting up in 'swift_cc_test' currently.

All the rest of the targets are getting compiled with -Wno-error=deprecated-declarations so the warnings are still there.

Maybe I should use a more descriptive name for this flag to make it more obvious it's only for tests?

PR #119, opened 2023-10-10
use lld on x86_64 darwin [DIP-358]

isaactorz:
Replaces `ld64`(linker) and `libtool`(archiver) with `lld` and `llvm-ar` on x86_64-darwin

This has a couple of advantages:
- x86_64 builds are now more hermetic and less susceptible to breakages from mac toolchain updates
- lld and ar are a lot less noisey and the resulting builds are astonishingly quiet.

This is not yet available on apple silicon. As of v14.0 of llvm - there are no prebuilt
binaries available for m1 and we are not buildling lld or other clang tools like clang-tidy in our custom builds.

It's worth mentioning that many organizations do not use lld for their production builds in favor of the system apple linker.

In our case we don't actually ship anything to apple - it's just a development platform - so I think it makes sense for us
to go full llvm for mac due to the above benefits.

I've tested a handful of build configurations on intel macs and m1 macs


cc/toolchains/llvm/swift_custom_features.bzl
```diff
@@ -88,6 +88,9 @@ swift_no_default_warnings = feature(
                 "-Wno-fortify-source",
                 "-Wno-absolute-value",
                 "-Wno-format",
+                "-Wno-deprecated-declarations",
+                "-Wno-unused-but-set-variable",
+                "-Wno-unused-variable",
```
pcrumley: Naïve question -- how are we supposed to know what warning can be removed here? Should we wait for someone from the platform team to ok it?

cc/toolchains/llvm/swift_custom_features.bzl
```diff
@@ -88,6 +88,10 @@ swift_no_default_warnings = feature(
                 "-Wno-fortify-source",
                 "-Wno-absolute-value",
                 "-Wno-format",
+                "-Wno-deprecated-declarations",
```
isaactorz, line 91: A few default compiler warnings I missed in the last PR.

Again these are only silenced for third party code (curl, suiteparse, hdf5, etc..)

pcrumley (commented):
My concern about this is that if lld is far quieter, could we be missing important warnings in the compilation process? It's true we don't ship linked binaries from MacOs, but we do use them internally, and I think its something where MacOS is a large fraction of the dev's OS. The warnings given to the devs on their local build might be the only place they could be reasonably surfaced and inspected (note: i admit some ignorance here, if in CI we have `-Wl,--fatal-warnings` flag or equivalent thing passed to the linker then this point is moot).

This comment also assumes the warnings actually have some value to them, if they don't then... it's just log spam that will cause more important logs to be missed.

cc/toolchains/llvm/swift_custom_features.bzl
pcrumley: you clarified while I was typing up my review.

isaactorz: These are all flags that are enabled by default in clang - so when we're building third party code (hdf5, suiteparse, curl, etc..) - the builds are spammed with a bunch of these warnings from code we don't really control.

The list here is only intended to be silenced in the third party code. In our code we error if these come up (`-Werror`).

@woodfell reviewed the last PR where I introduced this list to start removing noise from the builds. This is just adding on with a few I missed.

isaactorz:
> This comment also assumes the warnings actually have some value to them, if they don't then... it's just log spam that will cause more important logs to be missed.

Yeah it's pretty much exactly this. The native mac linker and archiver are warning about things that I would consider noise -
so it's distracting from more important messages.

This PR would a good read to go through about justification for a lot of the warning messages I've been suppressing:
https://github.com/swift-nav/rules_swiftnav/pull/118

And also good to keep in mind that the main benefit of doing this is getting to a hermetic toolchain on mac and all
of the associated benefits. The fact that lld is also quieter is sort of just an added bonus.

pcrumley (approved):
Lgtm but I would love @woodfell or similar to take a look at the additional warning suppressions.

woodfell (approved):

PR #122, opened 2023-10-20
Script to fix include guards [AP-243]

woodfell:


isaactorz (commented):
Thanks for adding this here Matt.

I just have one small request about the directory layout. I think things are going to get a bit untidy if we just add a new directory for every utility like this (and indeed there are already some "packages" I don't think should be hanging out at the top level).

I was thinking maybe we can put this an a package that lives in `cc/tools`?:
```
- cc/
  - tools/
    - BUILD.bazel
    - fix_include_guards.py
    - ...
```

What do you think about that?

woodfell:
> Thanks for adding this here Matt.
> 
> I just have one small request about the directory layout. I think things are going to get a bit untidy if we just add a new directory for every utility like this (and indeed there are already some "packages" I don't think should be hanging out at the top level).
> 
> I was thinking maybe we can put this an a package that lives in `cc/tools`?:
> 
> ```
> - cc/
>   - tools/
>     - BUILD.bazel
>     - fix_include_guards.py
>     - ...
> ```
> 
> What do you think about that?

Works for me

isaactorz (approved):

PR #123, opened 2023-10-20
fix: cc: toolchain: add sha256 for aarch64 llvm toolchain

isaactorz:
Adds the sha256 checksum to validate the llvm download.

Besides security - without the sha this artifact is not kept in bazels repository cache.


pcrumley (approved):
Whoops. Nice catch

PR #125, opened 2023-12-07
Use platforms v0.0.5

woodfell:
Resolves in incompatibility between the default version `0.0.3` and `@rules_pkg` which shows up in certain workspaces

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -16,11 +16,11 @@ load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
     name = "platforms",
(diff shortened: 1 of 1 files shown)


isaactorz (commented):
`WORKSPACE` isn't transitive

Every repo needs to provide the dependencies in their own `WORKSPACE` file.

FWIW bazel has rolled out a proper dependency management system that does what you're trying to do here: https://docs.bazel.build/versions/5.0.0/bzlmod.html#bazel-modules but this happened midway through our migration and much of the ecosystem didn't support it yet.

Updating our codebase to use this system is high on my priority list for next year.

PR #132, opened 2024-04-04
Remove unnecessary condition in _safe_flags function

reimerix:
## Description

Cover the following topics in the description:

- [ ] Rational
- [ ] Context
- [ ] Design notes

## Jira Reference

Link to a Jira ticket:

- [ ] https://swift-nav.atlassian.net/browse/BOARD-123

### DIFF
--- a/clang_tidy/clang_tidy.bzl
+++ b/clang_tidy/clang_tidy.bzl
@@ -126,7 +126,7 @@ def _safe_flags(flags):
         "-Wno-stringop-overflow",
     ]
 
-    return [flag for flag in flags if flag not in unsupported_flags and not flag.startswith("--sysroot") and not "-std=" in flag]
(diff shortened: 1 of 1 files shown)


isaactorz (approved):
I *think* we probably added this to work around the fact that bazel doesn't distinguish between options for c++ or c files, so we were running into issues where it would pass `-std=c++14` (or similiar), when compiling a c file, and that would error because we're building with `-Werror` (warning that the option is not recognized for c files).

But we figured out how to work around that quite a while ago so we probably just forgot to remove this hack.

Assuming this doesn't break anything I think we should definitely take this fix.

pcrumley (approved):
@isaactorz @swift-nav/internal-ops 

-- I think this is another place we should consider removing the devinfra team from CodeOwners

PR #134, opened 2024-05-16
feat: add rule for downloading private github assets

isaactorz:
Github release assets from private repositories are not available at the download link url like public assets.

They require a relatively complicated sequence of http requests to get the actual content - which is not possible with any of the OTB bazel donwloader rules (`http_file`, `http_archive`, etc..).

Luckily someone else has already solved this problem for us:
https://www.stevenengelhardt.com/2023/05/23/practical-bazel-downloading-private-release-assets-from-github/

I've taken this code so that we can download private github assets

Related: https://github.com/swift-nav/starling/pull/9180

### DIFF
--- a/tools/github_release_asset_file.bzl
+++ b/tools/github_release_asset_file.bzl
@@ -1,0 +1,72 @@
+# github_release_asset_file: Like http_file(), but works with GitHub
+# release assets.
+#
+# Created because downloading private release assets from GitHub is a PITA.
+
+load(":github_release_context.bzl", "github_release_context")
+
+def _impl(ctx):
+    downloaded_file_path = ctx.attr.downloaded_file_path if ctx.attr.downloaded_file_path else ctx.attr.asset_name
+
+    # Create file/ directory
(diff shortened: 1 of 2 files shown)


pcrumley (approved):
Is the best way to take in upstream functions like this copying and pasting them? 

Kind of feels like a step backwards when it comes to dependency management but I assume this is just commonplace in bazel

isaactorz:
> Is the best way to take in upstream functions like this copying and pasting them?
> 
> Kind of feels like a step backwards when it comes to dependency management but I assume this is just commonplace in bazel

It's just some guys blog post - there's a chance that this made it into one of the more popular upstream utility rules sets (like [bazel-lib](https://github.com/aspect-build/bazel-lib), or [skylib](https://github.com/bazelbuild/bazel-skylib)) but admittedly I did not check.

But yes in general there's a lot of copy pasting rules in the bazel ecosystem. The new modules system and the [Bazel Central Registry](https://registry.bazel.build/) aims to improve this situation but it's still pretty early.

PR #136, opened 2024-06-27
Extra string tags to image_stamp_tag, update codeowners[OC-1439]

john-michaelburke:
* Optionally add a list of extra string tags to either the image_tag or image_stamp_tag output.
* At least for my use case, I would like to use image_stamp_tag to tag not only the git version but add in a "latest-master" tag to the target image.
* I could see value in extending the image_stamp_tag to apply the regex functionality to extra tags, but I'll leave that up to a future user.
* As far as I understand, because I use a default value, this feature should be backwards compatible.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
-* @isaactorz
+* @pcrumley
+* @john-michaelburke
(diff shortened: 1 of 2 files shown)


pcrumley (approved):
this looks fine. Hopefully you are correct about it being backwards compatible.

john-michaelburke:
> this looks fine. Hopefully you are correct about it being backwards compatible.

Fortunately any users have to manually uprev rules_swiftnav so I think it is safe

PR #143, opened 2024-11-13
Update clang tool config files according to official guidelines [AP-3171]

sbmueller:
See https://swift-nav.atlassian.net/wiki/spaces/QM/pages/3059089594/Software+Coding+Guidelines


armallen (approved):
I see `starling`, `starling-core`, `orion-engine` and `orion` use local `.clang-format` files, and sometimes `.clang-tidy` too. Do you know which config will prevail when running the format/tidy bazel targets? 

Will this require integration work in `starling/starling-core` because of `readibility-identifier-naming`?

I ran `orion-engine`'s integrity tidy target locally with and without `orion-engine/.clang-tidy-integrity` and it's OK in both cases 👍 
```
bazel build --override_repository=rules_swiftnav=/workspaces/orion-engine/tmp/rules_swiftnav/ --config=clang-tidy-check-integrity //...
```

sbmueller:
@armallen I think there is a logic to take the local config files if they exist:
https://github.com/swift-nav/starling/blob/master/BUILD.bazel#L27

Good thing you point that out, that's actually not how it should be done, as this allows deviating from the coding guidelines without any sign-off.

sbmueller:
@armallen can you handle `orion` and `orion-engine` and I will create a ticket for the others?

armallen:
I'll handle the .clang-tidy-integrity file. 
Regarding .clang-format, I think removing the local file will have a negative impact on developers that rely on IDE auto formatting (which I think is most developers?).
I can remove it from the bazel call to make sure CI uses the correct file (therefore ensuring correct formatting according to `rules_swiftnav`), and align the existing ones with the one from `rules_swiftnav` to avoid any action from devs, WDYT?

scarcanague:
@armallen Is there additional action on Correction/Apps team / Orion repos owners on top of the .clang-tidy-integrity? 
@sbmueller  could you ping the #engineering channel as well for awareness if this may lead to CI failing for some repos pulling this file?

scarcanague (approved):

armallen:
@scarcanague  Yes I have a PR up for `orion-engine` evaluating impacts of these changes https://github.com/swift-nav/orion-engine/pull/8925
This will imply a reformat of orion-engine (and probably orion) sources, some minor code updates to comply to clang-tidy check.

PR #145, opened 2024-12-04
add mingw support [AP-2090]

sokhealy:
Add MingW Cross Compiler Toolchain

Testing: The current consumer of this toolchain is Starling-Core.
This toolchain is limited to an x86_64 Linux host to build an x86_64 Windows target. With this toolchain, we are generating a binary and a static library.

The artifacts were generated using CI/CD and were tested on Linux using Wine and on a Windows 11 system.
The static library that was generated was also tested on Windows 11 to generate the binary and static data was passed, and the results were compared.

The same steps above were also done on an x86_64 Ubuntu 24.04 developer machine.


WORKSPACE
```diff
@@ -0,0 +1 @@
+workspace(name = "rules_swiftnav")
```
sbmueller, line 1: Question: What's the difference between this and the `WORKSPACE.bazel` file, which contain the same line?

cc/toolchains/llvm_x86_64_windows/BUILD.bazel
```diff
@@ -0,0 +1,130 @@
...
+    srcs = [
+        ":compiler_files",
+        ":linker_files",
+        "@llvm_mingw_toolchain//:bin",
+    ],
+)
+
+config(name = "config")
+
+cc_toolchain(
```
sbmueller, line 101: Question: Comparing this with other toolchain files, I notice there is no call to `cc_toolchain_config`, why is that?

cc/toolchains/llvm_x86_64_windows/BUILD.bazel
sbmueller, line 1: Question: We already have an `llvm` folder in `toolchains`, why did you choose to put the llvm mingw files next to the llvm instead of a subfolder in `llvm`?

sbmueller (commented):
Difficult to review for me as I'm unfamiliar with Bazel. For me the most valuable information would be how this was tested, could you maybe extend the description by that?

cc/toolchains/llvm_x86_64_windows/BUILD.bazel
sokhealy: There is, 
https://github.com/swift-nav/rules_swiftnav/blob/sokhealy/mingw_support/cc/toolchains/llvm_x86_64_windows/config.bzl#L130-L142
I based this toolchain off of the other cross-compile toolchain that we are using.

WORKSPACE
sokhealy: It depends on the version we are using. This was added because I was experimenting with a different bazel version. I will remove. Thanks for catching this.

cc/toolchains/llvm_x86_64_windows/BUILD.bazel
sbmueller: Got it, thanks for the explanation!

sbmueller (approved):
Thanks for the clarifications, LGTM

Madhunika-atmakuri (approved):
From the parts I could understand, LGTM. We can merge this for now and come back to add additional stuff if needed.

PR #147, opened 2024-12-19
Fix bug that resulted in empty linker flags on MacOS [AP-3221]

sbmueller:
In `9b5d663`, a bug was introduced that resulted in empty linker flags
on MacOS. In Starlark, `[a] + [b] if c else []` results in `[]`, if `c` is False. To
just make `[b]` conditional, parenthesis must be used: `[a] + ([b] if c
else [])`.


armallen (approved):
Awesome!

IsakTjernberg (approved):

junc52 (approved):
Respect to find this!

PR #151, opened 2025-03-27
clang-tidy: Fix class member naming checks

sbmueller:
Our guidelines are, that public and protected class members are snake_case without suffix. Only private members have a `_` suffix.
Tested locally.


reimerix (approved):
LGTM, thank you!

armallen (approved):
🔥

PR #152, opened 2025-05-07
Update protobuf to v25.1

RReichert:
# Changes

I'm adding a new build flag which controls what C++ version it will use across the entire codebase, including third party libraries. By default Bazel uses C++14 (implicitly, even in third party libraries), which normally won't be a problem if it wasn't for the fact that the newest protobuf library which we are introducing didn't have have a problem with this, specifically Abseil.

Repositories can now run:

```sh
$ bazel build --@rules_swiftnav//:global_cxx_standard=17 //...
```

Which compiles everything with the C++17 standard. Options for C++20 and C++23 have been added as well.

### DIFF
--- a/cc/toolchains/llvm/cc_toolchain_config.bzl
+++ b/cc/toolchains/llvm/cc_toolchain_config.bzl
@@ -77,10 +77,12 @@ def cc_toolchain_config(
         "-fdata-sections",
     ]
 
-    cxx_flags = [
-        # The whole codebase should build with c++14
-        "-std=c++14",
-    ]
+    cxx_flags = select({
+        "//cc:global_cxx17": ["-std=c++17"],
+        "//cc:global_cxx20": ["-std=c++20"],
+        "//cc:global_cxx23": ["-std=c++23"],
+        "//conditions:default": ["-std=c++14"],
+    })
 
     link_flags = [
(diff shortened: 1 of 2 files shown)


cc/toolchains/llvm/cc_toolchain_config.bzl
```diff
@@ -77,10 +77,12 @@ def cc_toolchain_config(
         "-fdata-sections",
     ]
 
-    cxx_flags = [
-        # The whole codebase should build with c++14
-        "-std=c++14",
-    ]
+    cxx_flags = select({
```
sbmueller, line 80: How does this compare to https://github.com/swift-nav/rules_swiftnav/blob/main/cc/defs.bzl#L62 ? I actually wasn't aware of the flag directly in the toolchain file.

RReichert: what what is highlighted here, if you use the bazel `cc_library` or `cc_binary` directly it will impose the "global cxx standard". in the code snippet that [you posted](https://github.com/swift-nav/rules_swiftnav/blob/main/cc/defs.bzl#L62), that only applies to `swift_cc_library` and `swift_cc_binary` targets.

sbmueller (approved):

PR #154, opened 2025-05-22
Update bazel doxygen script [AP-3535]

richarddeurloo:
This PR support the templated Doxygen configuration files added to the hummingbird repo, which are used in starling-core. This PR adds support for a number of new templated variables.

Related PR:
- https://github.com/swift-nav/starling-core/pull/898
- https://github.com/swift-nav/hummingbird/pull/2

### DIFF
--- a/doxygen/doxygen.bzl
+++ b/doxygen/doxygen.bzl
@@ -8,8 +8,8 @@
 # EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
 # WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
 
-load("//tools:configure_file.bzl", "configure_file_impl")
 load("@rules_pkg//pkg:pkg.bzl", "pkg_zip")
+load("//tools:configure_file.bzl", "configure_file_impl")
 
 def _swift_doxygen_impl(ctx):
     vars = ctx.attr.vars | {}  # copy dict instead of referencing it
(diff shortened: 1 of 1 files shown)


doxygen/doxygen.bzl
```diff
@@ -18,37 +18,36 @@ def _swift_doxygen_impl(ctx):
...
         sed -i "s|@DOXYGEN_DOT_FOUND@|$DOXYGEN_DOT_FOUND|g" {config}
         sed -i "s|@DOXYGEN_DOT_PATH@|$DOXYGEN_DOT_PATH|g" {config}
         sed -i "s|@PLANTUML_JAR_PATH@|/usr/local/bin/plantuml.jar|g" {config}
-        sed -i "s|@INPUT_DIR@|{input_dir}|g" {config}
-        sed -i "s|@PROJECT_NAME@|{project_name}|g" {config}
+        sed -i "s|@INPUT_DIR@|$PROJECT_SOURCE_DIR|g" {config}
+        sed -i "s|@PROJECT_NAME@|$PROJECT_NAME|g" {config}
+        sed -i "s|@STABLE_GIT_TAG@|$STABLE_GIT_TAG|g" {config}
+        sed -i "s|@DOXYGEN_EXCLUDE@|$DOXYGEN_EXCLUDE|g" {config}
+        sed -i "s|@PROJECT_SOURCE_DIR@|$PRW|g" {config}
```
armallen, line 47: The following issue is not related to your PR, but to the way this script is implemented in the first place. I believe the doxygen actions are not correctly setup to be reproducible on macos.

I wanted to test this new version on integrity doxygens, and it turns out this sed command does not seem to be compatible with macos sed, which requires an argument to the `-i` option.

Using the following patch:
```
        platform=$(uname)
        if [ "$platform" = "Darwin" ]; then
            SED='sed -i ""'
        else
            SED="sed -i "
        fi
        // Updated sed calls
        $SED "s|@DOXYGEN_DOT_FOUND@|$DOXYGEN_DOT_FOUND|g" $config
```

I then get `in-place editing only works for regular files` because the doxygen files are links, not regular files.

So I updated the calls again to:

```
        config=$(readlink -f {config})
        # backward compatibility with old CMake-style doxygen config files
        $SED "s|@DOXYGEN_DOT_FOUND@|$DOXYGEN_DOT_FOUND|g" $config
```

Then doxygen is not found (I have it installed, but PATH seems to be cleared when running this script).

```
/bin/bash: line 28: doxygen: command not found
// $PATH value is the following, bazel probably isolates it from my user PATH
/usr/gnu/bin:/usr/local/bin:/bin:/usr/bin:.
```

armallen: @richarddeurloo  @sbmueller  Then my question is, are you able to build doxygens on your mac? Should we find a way to make this script platform independent, and add doxygen binaries as explicit dependencies?

richarddeurloo: @armallen, I agree and this script is a thorn in my eye as well. I don't have time to clean this up now. I just wanted to "make it work" for the moment. We have another ticket to deal with the refactor.

And you are right about the `sed` and `doxygen` issues on Mac. This is an existing problem from the script. The only way around it for the moment is to run it in a docker container on Mac. I've tested it and it works. See my comment in https://github.com/swift-nav/starling-core/pull/898: For bazel: works-in-local-docker™️. 😬

doxygen/doxygen.bzl
```diff
@@ -18,37 +18,36 @@ def _swift_doxygen_impl(ctx):
...
     config = configure_file_impl(ctx, vars, ctx.attr.name + "_Doxyfile")[0].files.to_list()[0]
 
-    input_dir = ""
-    if "PROJECT_SOURCE_DIR" in vars:
-        input_dir = vars["PROJECT_SOURCE_DIR"]
-
-    project_name = ""
-    if "PROJECT_NAME" in vars:
-        project_name = vars["PROJECT_NAME"]
-
```
richarddeurloo, line 30: Note: these are now directly exposed as env variable in bash by setting `env = vars` below.

doxygen/doxygen.bzl
sbmueller: @armallen I've stumbled upon the exact same issues. I took a look a few weeks ago (see my comments in above ticket) and I can't help but realize the doxygen bazel implementation is a mere PoC for Linux which I think never was intended as a final and general solution. This definitely needs work.

doxygen/doxygen.bzl
```diff
@@ -18,37 +18,36 @@ def _swift_doxygen_impl(ctx):
...
         if command -v dot &> /dev/null
         then
             DOXYGEN_DOT_FOUND=YES
             DOXYGEN_DOT_PATH=$(which dot)
         fi
 
+        PRW=`pwd`
+
+        # backward compatibility with old CMake-style doxygen config files
```
sbmueller, line 39: Do we still use old doxygen config files?

sbmueller (approved):
:shipit:

doxygen/doxygen.bzl
richarddeurloo: What I meant here was the use of `@VAR@` in the config files. Unfortunately, I had to maintain those in the new Doxygen config files. Otherwise the CMake builds would fail. So I was either maintain it here (for the moment) or have duplicate config files, for CMake and Bazel. That didn't seem like the best approach. I hope we can drop CMake at some point and use the default Doxygen approach `$(VAR)`.

PR #156, opened 2025-06-27
Use labels to work properly as external dependency [IO-410]

martin4861:
In order to properly consume `rules_swiftnav` with as a bazel module, labels need to be used with `@`.

This change is backwards compatible and works with the `WORKSPACE` file.


RReichert (approved):
make sure you verify these changes will work with orion and orion-engine.

armallen (approved):

martin4861:
> make sure you verify these changes will work with orion and orion-engine.

Tests passed in 
✅ starling-core
✅ orion
✅ orion-engine

PR #162, opened 2025-07-07
Implement MODULE [IO-410]

martin4861:
Implementing the `MODULE.bazel` file for the usage with `bzl_mod`.


MODULE.bazel
```diff
@@ -0,0 +1,49 @@
...
+    name = "rules_swiftnav",
+    version = "0.1.0",
+    compatibility_level = 1,
+)
+
+# Core Bazel dependencies
+bazel_dep(name = "bazel_skylib", version = "1.4.1")
+bazel_dep(name = "platforms", version = "0.0.6")
+bazel_dep(name = "rules_cc", version = "0.0.9")
+bazel_dep(name = "rules_pkg", version = "0.9.1")
```
armallen, line 11: Out of curiosity: how did you choose these versions? Which versions were we using before when not using bzlmod?

MODULE.bazel
```diff
@@ -0,0 +1,49 @@
...
+    "@rules_swiftnav//cc/toolchains/llvm/x86_64-linux:cc-toolchain-x86_64-linux",
+    "@rules_swiftnav//cc/toolchains/llvm/x86_64-linux:cc-toolchain-intel-mkl",
+    "@rules_swiftnav//cc/toolchains/llvm/x86_64-aarch64-linux:cc-toolchain-aarch64-bullseye-graviton2",
+    "@rules_swiftnav//cc/toolchains/llvm/x86_64-aarch64-linux:cc-toolchain-aarch64-bullseye-graviton3",
+    "@rules_swiftnav//cc/toolchains/musl/aarch64:toolchain",
+    "@rules_swiftnav//cc/toolchains/musl/armhf:toolchain",
+    "@rules_swiftnav//cc/toolchains/musl/x86_64:toolchain",
+    "@rules_swiftnav//cc/toolchains/llvm_x86_64_windows:mingw_toolchain",
+    "@rules_swiftnav//cc/toolchains/gcc_arm_embedded:toolchain",
+    "@rules_swiftnav//cc/toolchains/yocto_generic:toolchain",
```
armallen, line 45: How many of these toolchains are used `starling-core` CI?

armallen (commented):
I'm not too familiar with bzlmod so I have lots of questions 😄
* Did you test this locally, or in CI in some `starling-core` branch with updated bazel version?
* Does using `bzlmod` in some bazel dependency implies using it in all the build graph, or can we propagate it iteratively?
* `rules_swiftnav` now has both WORKSPACE and MODULE files, which one is used by default?

MODULE.bazel
martin4861: With the `WORKSPACE` files function calls like [`go_rules_dependencies`](https://github.com/bazel-contrib/rules_go/blob/release-0.30/go/private/repositories.bzl#L23) bring in the dependencies to e.g. `platforms` (0.4.0), `bazel_skylib` (1.1.1).
These dependencies are actually direct dependencies of `rules_swiftnav`, therefore they need to be declared in the `MODULE.bazel` file.
I just took some versions, which are a bit newer and which seem to work, when using this repo.

If somebody uses `rules_swiftnav` downstream those versions can be easily overwritten by the [version selection](https://bazel.build/external/module#version-selection). To ensure reproducibility a `MODULE.lock` file is created to pin the resolved versions, see the integration [PR](https://github.com/swift-nav/starling-core/pull/983/files#diff-ae888452fd6b0b69f2f29dd1e29430c0f3f59f8b68b313f120b62b855c5f9fe0).

armallen (approved):

PR #166, opened 2025-07-24
Implement swift_c_test

martin4861:
Implement `swift_c_test`, which has the same behavior as `swift_cc_test`, except for setting the `cpp` standard.


armallen (approved):
Awesome!

PR #188, opened 2025-09-16
have cc_static_library return merged CcInfo

notoriaga:
In order to use the static lib in other bazel targets we need this to return CcInfo. This is useful for rust where I can create a static version of orion_proto, then use that in a new cc_library that adds in some rust specific code.


martin4861 (approved):
lgtm!

PR #192, opened 2025-09-25
Add standard argument to `swift_cc_test_library` and `swift_c_test`

sbmueller:
`swift_cc_test_library` and `swift_c_test` did not support the `standard` attribute, which allows an explicit selection of the language standard. We want to be able to precisely set the C/C++ standard for all targets. All other swift related rules support this already.

Reviving @armallen's old commits.

### DIFF
--- a/cc/defs.bzl
+++ b/cc/defs.bzl
@@ -592,14 +592,15 @@ def swift_cc_test_library(**kwargs):
             be relative to the package this macro is called from.
     """
 
-    _ = kwargs.pop("nocopts", [])  # To handle API compatibility.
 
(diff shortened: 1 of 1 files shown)


cc/defs.bzl
```diff
@@ -592,11 +592,15 @@ def swift_cc_test_library(**kwargs):
             be relative to the package this macro is called from.
     """
 
-    _ = kwargs.pop("nocopts", [])  # To handle API compatibility.
-
+    nocopts = kwargs.pop("nocopts", [])  # pop because nocopts is a deprecated cc* attr.
```
armallen, line 595: Had a look at the rest of the file, and it seems `swift_c_test`  might also be missing the `standard` option? `swift_cc_test` uses it.

armallen (approved):

PR #236, opened 2026-04-15
Implement tooling to merge SARIF reports

martin4861:
Add tooling, which reads SARIF reports from `rules_lint` and merges them, so that they can be easily processed by tooling like reviewdog.

### DIFF
--- a/tools/lint/extract_lint_results.py
+++ b/tools/lint/extract_lint_results.py
@@ -1,0 +1,511 @@
+#!/usr/bin/env python3
+"""
+Extract and merge linting results from a Bazel build.
+
+Collects SARIF (Static Analysis Results Interchange Format) report files and
+patch files from a Bazel Build Event Protocol JSON file, merges the SARIF
+reports into a single file for efficient processing by reviewdog, and copies
+non-empty patch files to a specified directory.
+"""
+
+import argparse
+import json
+import os
+import re
+import shutil
+import sys
+from pathlib import Path
+from typing import Any
+
+
+def extract_files_from_bep(
+    build_event_json_file: Path,
+    bazel_output_path: Path,
+    ext: str,
+) -> list[Path]:
+    """
+    Extract file paths with the given extension from a Bazel Build Event Protocol JSON file.
+
+    The BEP file is newline-delimited JSON. Each line is parsed for namedSetOfFiles
+    entries, and files whose name ends with the given extension are collected.
+
+    Args:
+        build_event_json_file: Path to the Bazel build event JSON file
+        ext: File extension to filter
+        bazel_output_path: Workspace root used to resolve relative BEP paths
+            (e.g. paths starting with bazel-out/). When omitted, paths are kept
+            as-is (relative to the current working directory).
+
+    Returns:
+        List of matching file paths
+    """
+    report_files = []
+
+    with open(build_event_json_file, "r") as f:
+        for line in f:
+            line = line.strip().rstrip("\r")
+            if not line:
+                continue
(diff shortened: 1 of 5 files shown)


tools/lint/extract_lint_results.py
```diff
@@ -0,0 +1,420 @@
...
+        workspace_root: Repository root used to resolve _virtual_includes symlinks
+
+    Returns:
+        Normalized path relative to repository root
+    """
+    # Pattern to match Bazel execroot paths like:
+    # ../../../902/execroot/_main/pvt_engine/...
+    # ../../execroot/_main/pvt_engine/...
+    # The pattern captures everything after 'execroot/_main/' or 'execroot/__main__/'
+    execroot_pattern = r"(?:\.\./)*(?:\d+/)?execroot/(?:_main|__main__)/(.*)"
```
reimerix, line 93: Should there be a test that we can use to check compatibility with new bazel versions, or is this path going to be stable?

tools/lint/extract_lint_results.py
```diff
@@ -0,0 +1,420 @@
...
+    # resulting path matches the PR diff (e.g. config_utils/include/... instead of
+    # bazel-out/.../config_utils/_virtual_includes/...).
+    if "_virtual_includes" in normalized and workspace_root is not None:
+        full_path = workspace_root / normalized
+        try:
+            resolved = Path(os.path.realpath(full_path))
+            canonical_root = Path(os.path.realpath(workspace_root))
+            normalized = str(resolved.relative_to(canonical_root))
+        except (OSError, ValueError):
+            pass  # Keep original if the symlink cannot be resolved
```
reimerix, line 115: Suggestion for future: if need arises, we should consider splitting this out in a separate package to avoid duplication this seems like it could be useful for a number of use cases.

reimerix (approved):

tools/lint/extract_lint_results.py
```diff
@@ -0,0 +1,420 @@
...
+
+    with open(build_event_json_file, "r") as f:
+        for line in f:
+            line = line.strip().rstrip("\r")
+            if not line:
+                continue
+            try:
+                event = json.loads(line)
+            except json.JSONDecodeError:
+                continue
```
armallen: Is this a real possibility or just an llm generated try/except? I often see json.loads calls surrounded with a try/except block, which can make sense if we expect some incorrect data, but I feel like we expect only valid data in the build_event_json, and that an incorrect json content would entail a real issue wich should be raised at a higher level. 

If that's the case then we can remove the try/except and let python raise the exception and the script fail.

tools/lint/extract_lint_results.py
```diff
@@ -0,0 +1,420 @@
...
+                    merged_runs.append(normalized_run)
+                    normalized_paths += len(results)
+                processed_files += 1
+
+        except json.JSONDecodeError as e:
+            print(f"Error parsing {input_file}: {e}", file=sys.stderr)
+            continue
+        except Exception as e:
+            print(f"Error processing {input_file}: {e}", file=sys.stderr)
+            continue
```
armallen: Same here, do we expect incorrect JSON content in sarif reports? I feel like the script should fail in that case, as it would report incomplete data

tools/lint/extract_lint_results.py
martin4861: That's a valid point, however it's not yet straight forward to test against newer versions of Bazel, because some dependencies are not yet fully compatible with Bazel 9. If it breaks, we have to re-iterate.

PR #241, opened 2026-05-20
Implement hermetic doxygen

martin4861:
Implement hermetic doxygen which allows out-of-the-box execution on both Linux and Mac.
This additionally adds a dependency of the generated doxygen output to the doxygen executables so that version updates of doxygen force a rebuild of the documentation.

### DIFF
--- a/doxygen/extensions.bzl
+++ b/doxygen/extensions.bzl
@@ -1,0 +1,40 @@
+# Copyright (C) 2026 Swift Navigation Inc.
+# Contact: Swift Navigation <dev@swift-nav.com>
+#
+# This source is subject to the license found in the file 'LICENSE' which must
+# be be distributed together with this source. All other rights reserved.
+
+"""Module extension that fetches hermetic Doxygen binaries from upstream GitHub releases."""
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+_DOXYGEN_VERSION = "1.17.0"
+_DOXYGEN_RELEASE = "Release_1_17_0"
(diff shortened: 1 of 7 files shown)


martin4861:
@claude please review

doxygen/extensions.bzl
```diff
@@ -0,0 +1,40 @@
...
+    "doxygen_macos_arm64": struct(
+        asset = "doxygen-{v}-mac-arm.zip".format(v = _DOXYGEN_VERSION),
+        strip_prefix = "doxygen-{v}".format(v = _DOXYGEN_VERSION),
+        build_file = "@rules_swiftnav//doxygen:doxygen_macos.BUILD",
+        sha256 = "e05a7f647f894d2d82ef26a1d231cda079d2d1d85cf1e1c6fd8072430d80d614",
+    ),
+}
+
+def _doxygen_extension_impl(_ctx):
+    for repo_name, spec in _DOXYGEN_ARCHIVES.items():
```
sbmueller, line 31: Q: Will this download both archives? It would be nice to only download what is needed.

sbmueller (approved):
🚀 
Claude doesn't work here as it was not set up as GH action ([see](https://code.claude.com/docs/en/github-actions)).

doxygen/extensions.bzl
martin4861: This file declares the `http_archive`s, but the actual download is only invoked when something in the build graph dereferences the corresponding label. This happens in doxygen/BUILD.bazel, where we `select` based on the platform.

PR #242, opened 2026-06-16
Resolve strip tool via toolchain action config

pcrumley:
## Problem

The `strip` rule (`//cc:strip.bzl`) resolves its tool from `cc_toolchain.strip_executable` — the **legacy** `ToolchainInfo` field. Toolchains built on the modern rules_cc API (the swift `*-llvm20` toolchains, and the rules_rs hermetic LLVM toolchain) only wire the `strip` action through the action config and leave `strip_executable` pointing at a path that is never materialized. The action then fails at execution:

```
external/llvm+/toolchain/strip -S -p -o ... --strip-unneeded
execvp(external/llvm+/toolchain/strip, ...): No such file or directory
```

Toolchains that still set `strip_executable` (e.g. the musl SDK toolchains) work, which is why this only surfaced when consuming the LLVM toolchain (in `bird-sweater`'s rules_rs migration).

Separately, the rule only ever worked on Linux/ELF: it hardcodes `-p` and consumers pass `--strip-unneeded`, both of which are **ELF-only** options that `llvm-strip` rejects for Mach-O (`option is not supported for MachO`). `bird-sweater` needs to build on macOS as well as Linux.

## Fix

**Tool resolution.** Resolve the tool with `cc_common.get_tool_for_action(ACTION_NAMES.strip)` against a `configure_features` feature configuration. This reads the toolchain's action config and works for **both** the legacy `tool_paths`-based toolchains and the modern action-config toolchains (it returns `llvm-strip` for the LLVM toolchains and the SDK `*-strip` for the musl toolchains).

**macOS / Mach-O.** Make the flags platform-aware (still using the toolchain's `llvm-strip` — no switch to rustc strip):
- **Linux/ELF: byte-for-byte unchanged** — `llvm-strip -S -p -o OUT IN --strip-unneeded`.
- **macOS/Mach-O:** drop `-p` (preserve-dates; no symbol-stripping effect), and translate `--strip-unneeded` → `-x` so local/internal symbols are still stripped. Stripping protects shipped symbols, so we keep parity of *intent* rather than silently stripping less on macOS.

Consumers (bird-sweater) need no BUILD changes — the rule handles the platform difference.

## Test

`examples/small_world/src/strip_tool` — a `cc_binary` + `strip` + `build_test`, mirroring bird-sweater's usage (`stripopts = ["--strip-unneeded"]`). The example uses the swift LLVM toolchains (modern API), so it exercises exactly the path that was broken.

- The existing **ubuntu** job builds it via `bazel coverage //...` (ELF tool-resolution guard).
- A new **macos-latest** job runs `bazel test //src/strip_tool/...` (Mach-O flag guard). The ubuntu-only setup could not catch a Mach-O-only flag regression — which is how a rules_swiftnav change can silently break a downstream macOS build.

Verified locally on macOS: the strip action runs (`llvm-strip -S -o OUT IN -x`) and the stripped binary executes.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -38,3 +38,24 @@ jobs:
           cd examples/small_world
           bazel run //tools/buildifier:buildifier
           git diff --exit-code
+
+  strip-macos:
+    # Guards the strip rule against Mach-O regressions: llvm-strip rejects
+    # several ELF-only options on macOS, so this exercises the same strip()
+    # target that bird-sweater relies on. The Linux job above cannot catch a
+    # Mach-O-only flag regression.
+    runs-on: macos-latest
+    steps:
+      - name: Checkout source
+        uses: actions/checkout@v6
+
+      - uses: bazel-contrib/setup-bazel@0.19.0
+        with:
+          bazelisk-cache: true
+          disk-cache: ${{ github.workflow }}
+          repository-cache: true
+
+      - name: Build and test strip example
+        run: |
+          cd examples/small_world
+          bazel test //src/strip_tool/...
--- a/cc/strip.bzl
+++ b/cc/strip.bzl
@@ -2,10 +2,31 @@ load("@bazel_tools//tools/build_defs/cc:action_names.bzl", "ACTION_NAMES")
 load("@bazel_tools//tools/cpp:toolchain_utils.bzl", "find_cpp_toolchain")
 load("@rules_cc//cc/common:cc_common.bzl", "cc_common")
 
+# llvm-strip rejects these ELF-only options when operating on a Mach-O object
+# ("option is not supported for MachO"). They have no symbol-stripping effect
+# worth preserving, so drop them on macOS.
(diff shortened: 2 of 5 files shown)


martin4861 (commented):
I tried on Mac, but `bazel test //src/strip_tool/...` did fail with `llvm-strip: error: option is not supported for MachO`.

If it's just about stripping the binary, can we use the [built-in `binary.stripped`](https://github.com/bazelbuild/rules_cc/blob/f65555ccb4257d65e0a5379f9f476dd748203d4b/cc/private/rules_impl/cc_binary.bzl#L42-L50) target from `rules_cc`?
Following test worked for me on Mac
```
build_test(
    name = "strip_test2",
    targets = [":hello.stripped"],
)
```

pcrumley:
@martin4861 my understanding that the builtin in strip only works for cc targets. I want to use this for targets built with other toolchains like rust. I fixed the MacOS build

martin4861 (approved):

PR #244, opened 2026-06-23
Add automated release workflow

pcrumley:
## What

Adds a near-one-command release flow for `rules_swiftnav`.

**`bazel run //tools:release -- major|minor|patch`** — bumps `version` in
`MODULE.bazel`, refreshes `examples/small_world/MODULE.bazel.lock`, and opens a
release PR. Supports `--dry-run`. `--current` prints the current version and is
reused by CI.

**`.github/workflows/release.yaml`** — on push to `main`, detects the
`MODULE.bazel` version change (vs `github.event.before`) and creates the
`vX.Y.Z` tag + GitHub release. Uses only the built-in `GITHUB_TOKEN`,
fork-guarded, and idempotent (skips if unchanged or if the tag already exists).

Publishing the module to the downstream Bazel registry remains a separate manual
step — see `docs/RELEASING.md`.

## Notes

- `tools/release.py` is stdlib-only so CI runs it with plain `python3`.
- The version regex is anchored to the `module()` block so `bazel_dep` versions
  are never touched (covered by `//tools:release_test`).
- See the `GITHUB_TOKEN` breadcrumb comment in the workflow: events it creates
  won't trigger other workflows; switch to a GitHub App token if that's ever
  needed.

## Test plan

- [x] `bazel test //tools:release_test` (version parse/bump/set, --current)
- [x] `bazel run //tools:release -- minor --dry-run` → `0.20.0 -> 0.21.0`
- [x] Detection logic verified on real history (0.19.0→0.20.0 releases; unchanged commit skips; existing-tag guard skips)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,64 @@
+name: release
+
+# Cuts a release when MODULE.bazel's version changes on a push to main.
+#
+# NOTE: this job creates the tag + GitHub release using the built-in
+# GITHUB_TOKEN. Events created with GITHUB_TOKEN do NOT trigger other
+# workflow runs (GitHub's recursion guard). That is fine today: nothing
+# in this repo listens for the tag/release event, and downstream
+# publishing is handled out-of-band by a separate manual process. If a
+# future workflow is added `on: release` or `on: push: tags` (e.g. to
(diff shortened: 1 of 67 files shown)


martin4861 (approved):
That's cool, thanks! 🚀

PR #248, opened 2026-06-30
Add lint script to work with rules_lint

martin4861:
Extract `lint.sh` to reduce code duplication


armallen (approved):
Happy to see that! 🔥

PR #251, opened 2026-07-13
Add flag use_hermetic_llvm

martin4861:
Add flag `use_hermetic_llvm` to switch eigen to use the hermetic `@llvm-project//openmp:libomp` from hermetic llvm

### DIFF
--- a/third_party/BUILD.bazel
+++ b/third_party/BUILD.bazel
@@ -30,10 +30,7 @@ config_setting(
     visibility = ["//visibility:public"],
 )
 
-# Link MKL against the hermetic LLVM OpenMP runtime instead of the system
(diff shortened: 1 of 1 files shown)


third_party/BUILD.bazel
```diff
@@ -29,3 +29,22 @@ config_setting(
     flag_values = {":enable_mkl": "true"},
     visibility = ["//visibility:public"],
 )
+
+# Link MKL against the hermetic LLVM OpenMP runtime instead of the system
```
sbmueller: S: I wouldn't mention the MKL / openmp details here. The flag is not about MKL, but generally enables the hermetic LLVM.

sbmueller (approved):

PR #252, opened 2026-07-13
Make work with `rules_rs` bindgen

martin4861:
Append the clang resource folder at the end of every compilation, so that `rules_rs` hermetic `bindgen` can find builtin/resource headers. (rules_rs disables automatic search of headers, therefore we need to specify it, when using it with our own toolchains. Needed as long as we did not migrate to hermetic-llvm)

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -10,7 +10,7 @@
 
 module(
     name = "rules_swiftnav",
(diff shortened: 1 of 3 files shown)


pcrumley (approved):

pcrumley (approved):
:shipit:

PR #256, opened 2026-07-30
Add valgrind macros and reporting tooling

martin4861:
Implement valgrind macros for callgrind, memcheck and massif (incl. DRD). These macros use a python based runner script, which execute the binary using valgrind. Valgrind itself needs to be installed on the system and therefore those actions are not hermetic.
After having executed valgrind, each macros will then run the report generation, which extract the necessary metrics into a readable json file (see `//tools/valgrind/report`). Moreover the report tooling performs also a comparison against a baseline to block regressions found during profiling.

`//tools/valgrind/report:pr_comment` creates a markdown formatted output, which can be used in a Github PR comment.


tools/valgrind/report/callgrind_report.py
```diff
@@ -0,0 +1,164 @@
...
+_SUMMARY_RE = re.compile(r"^summary:\s*(\d+)")
+_TOTALS_RE = re.compile(r"^totals:\s*(\d+)")
+
+
+def find_output_files(output_dir: str) -> list[str]:
+    """Return the per-process callgrind dumps in output_dir, sorted by name."""
+    return sorted(
+        os.path.join(output_dir, name)
+        for name in os.listdir(output_dir)
+        if _OUTPUT_FILE_RE.match(name)
```
armallen: I noticed this file uses the old `os.path` / `str` approach to handle files, while the next one uses `Path`.
I'm a big advocate of using `Path` when we can!

tools/valgrind/report/callgrind_report_test.py
```diff
@@ -0,0 +1,165 @@
...
+            self.assertEqual(report["label"], "my_target")
+            self.assertEqual(report["baseline_file"], "pkg/baseline.json")
+            self.assertEqual(len(report["metrics"]), 1)
+            metric = report["metrics"][0]
+            self.assertEqual(metric["key"], "cpu_instructions")
+            self.assertEqual(metric["name"], "CPU instructions")
+            self.assertEqual(metric["unit"], "")
+            self.assertEqual(metric["value"], 1000000)
+            self.assertEqual(metric["baseline"], 980000)
+            self.assertEqual(metric["tolerance_pct"], 5.0)
```
armallen: For this line and all similar lines in the PR:


```suggestion
            self.assertAlmostEqual(metric["tolerance_pct"], 5.0)
```

tools/valgrind/report/callgrind_report_test.py
```diff
@@ -0,0 +1,165 @@
...
+    path.write_text(content)
+    return path
+
+
+class ReadCallgrindOutputTest(unittest.TestCase):
+    def test_sums_both_dump_formats_across_processes(self) -> None:
+        with tempfile.TemporaryDirectory() as tmp:
+            _write(Path(tmp) / "valgrind-callgrind.1", _SUMMARY_DUMP)
+            _write(Path(tmp) / "valgrind-callgrind.2", _TOTALS_DUMP)
+            self.assertEqual(sum_instructions(find_output_files(tmp)), 1250000)
```
armallen: I'm more familiar with pytest, but I believe this is not the standard way to handle fixtures. Unittest seems to be providing https://docs.python.org/3/library/unittest.html#unittest.TestCase.enterContext to handle temporary file fixtures. 

Not sure we need to update the tests really, as this python code lives in rules_swiftnav

tools/valgrind/report/callgrind_report_test.py
```diff
@@ -0,0 +1,165 @@
...
+                "--label",
+                "my_target",
+                "--baseline",
+                str(baseline),
+                "--baseline-label",
+                "pkg/baseline.json",
+                "--tolerance-pct",
+                tolerance_pct,
+            ]
+        )
```
armallen: Also a side note that does not imply code update here: we usually use main to do two things:
* setting up argparser (or `typer`, or any other cli handling lib)
* setting up logging
* calling the modules entrypoint function that has typed arguments

This way, we don't need to setup input arguments as a list of string and unit test only the entrypoint function. Then the script's main interface is only tested in "integration" tests.

tools/valgrind/runner.py
```diff
@@ -0,0 +1,237 @@
...
+        )
+        return 1
+
+    separator = argv.index("--")
+    opts = parse(argv[:separator])
+    try:
+        args = expand(argv[separator + 1 :])
+    except ValueError as err:
+        print(f"Error: {err}", file=sys.stderr)
+        return 1
```
armallen: I don't think this error should be caught if we only print it, it is printed when not caught too and return code is nonzero then.

tools/valgrind/runner.py
```diff
@@ -0,0 +1,237 @@
...
+    than asking massif for stacks. The report file is only written once there is
+    something in it, so a pass that produced nothing leaves the reporting tool
+    free to fall back on massif's own figure rather than reading an empty file.
+    """
+    log = SCRATCH_DIR / "drd.log"
+    drd = [
+        "--tool=drd",
+        "--show-stack-usage=yes",
+        "--first-race-only=yes",
+        f"--log-file={log}",
```
armallen, line 115: Claude comment:

```
DRD log is clobbered under --trace-children (runner.py, drd_stack_usage)
--log-file={log} has no %p, so with trace_children = True every child process opens and truncates the same drd.log. Stack-usage lines from all but the last-writing process are lost, silently under-reporting memory_stack_mb. Massif and callgrind both use %p in their out-files; the DRD log should too (then glob the logs).
```

which seems to be a valid comment, valgrind docs mention that %p should be used when trace-children is used.

https://valgrind.org/docs/manual/manual-core.html#manual-core.suppress:~:text=%25p%20is%20replaced%20with%20the%20current%20process%20ID

tools/valgrind/report/callgrind_report.py
```diff
@@ -0,0 +1,164 @@
...
+
+INSTRUCTIONS_FILENAME = "valgrind-callgrind.instructions"
+REPORT_JSON_FILENAME = "valgrind-callgrind.report.json"
+
+METRIC_KEY = "cpu_instructions"
+METRIC_NAME = "CPU instructions"
+
+# Only the per-process dumps, so the report files this tool writes into the
+# same directory are never mistaken for callgrind output on a re-run.
+_OUTPUT_FILE_RE = re.compile(r"^valgrind-callgrind\.\d+$")
```
armallen: Claude comment again here: 

_OUTPUT_FILE_RE = ^valgrind-callgrind\.\d+$ only matches .<pid>. With --separate-threads=yes (.<pid>-<tid>) or any multi-part dump options (.<pid>.<part>) in valgrind_args, those files don't match and their instructions vanish from the total

armallen (commented):
Awesome PR, this will be VERY useful. I mostly have python syntax comment

tools/valgrind/report/callgrind_report_test.py
martin4861: Makes sense. I'll refactor that

JADC362 (approved):

armallen (approved):

